PR #11, opened 2015-08-20
Playground to Master

Snugug:
Let's start deving for realz


Snugug:
That worked, but didn't close!

attiks:
Looks good, is ``` var gonzales = require('gonzales-pe') ``` needed in each rule.

I think the current approach of looping everything once is sane, can't think of a benefit if you loop for each rule separately.

Snugug:
@attiks No, not needed.

Switched to having each rule loop its own AST. For some more complex rules it's required.

PR #22, opened 2015-08-27
Adds a rule for !important

### LINKED ISSUE #21: New rule: !important
Add a rule that flags up the use of `!important` :fire:

bgriffith:
This adds a basic rule to detect the use of !important. By default it discourages it as this is well known bad practice. Fixes #21.


Snugug:
* Please rename the rule to `no-important`
* Please use the [add unique](https://github.com/sasstools/sass-lint/blob/master/lib/rules/indentation.js#L43-L49) pattern for adding a detect

Snugug:
And thanks for the contribution!

bgriffith:
No problem, I've updated. If there's anything else, let me know :+1:

Snugug:
Awesome, thanks!

PR #74, opened 2015-09-01
Add hex length rule

### LINKED ISSUE #73: New Rule: Hex Length
Hex length rule

as per [SCSS Lint](https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#hexlength)

DanPurdy:
Add a rule to enforce the length of hexadecimal values

Closes #73

### DIFF
--- a/docs/rules/hex-length.md
+++ b/docs/rules/hex-length.md
@@ -35,3 +35,13 @@ $foo-color: #456456;
   color: #ffffff;
 }
 ```
+
+In both cases the following will be allowed as the values cannot be shortened:
+
+```scss
+$quz-color: #abcdef
+
+.qux {
+  color: #123456
+}
+```
--- a/docs/rules/hex-length.md
+++ b/docs/rules/hex-length.md
@@ -1,0 +1,37 @@
+# Hex Length
+
+Rule `hex-length` will enforce the length of hexadecimal values
+
+## Options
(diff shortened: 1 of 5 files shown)


DanPurdy:
Ooops on the warning message I added.. changing now.

Snugug:
How does this work if a hex code can't be shortened? If it cannot be shortened, will it throw an error? If not, that should be included in the docs.

DanPurdy:
That's a very good point, I'll look into that now. it'll just need to check to make sure the first 3 values aren't equal to the last 3, should make it more robust either way.

DanPurdy:
I've added a separate block to the docs as the example would be correct in both instances.

Also checking whether values can be shortened or not when the style option is set to short.

Snugug:
Does the `color` type only return hex values, or does it return named color values as well? If it returns named color values, this will error.

DanPurdy:
I did check, it only returns hex values strangely.. 

I'll add some test cases though so we can make sure this doesn't break with updates to gonzales though.

Snugug:
Perfect, thanks

DanPurdy:
That should cover the eventualities, I do find it strange that they aren't returned currently though.

Snugug:
All looks good. Thanks!

PR #78, opened 2015-09-02
Add hex notation rule

### LINKED ISSUE #77: New Rule: Hex Notation
Check hexadecimal values are either all lowercase (default) or all uppercase.

as per [SCSS Lint](https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#hexnotation)

DanPurdy:
Add a rule to enforce the consistency of hexadecimal value cases.

Add 3 new helper functions
* isNumber - check is integer
* isUpperCase - check if string passed is all uppercase
* isLowerCase - check if string passed is all lower case

I kept the case helper functions separate to allow use in other rules to follow (property name case checking?). They allow you to pass many non alphanumeric characters too as part of a string without erroneously flagging them as upper or lower case.

Would like to add tests for all the helper functions too in the future.

Closes #77

### DIFF
--- a/docs/rules/hex-notation.md
+++ b/docs/rules/hex-notation.md
@@ -1,0 +1,45 @@
+# Hex Notation
+
+Rule `hex-notation` will enforce the case of hexadecimal values
+
+## Options
+
+* `style`: `lowercase`/`uppercase` (defaults to `lowercase`)
+
+## Examples
+
+When `style: lowercase`, the following are allowed. When `style: uppercase`, the following are disallowed:
+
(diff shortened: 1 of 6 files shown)


Snugug:

DanPurdy:
Done though, I'm guessing we may see these again once the next PR gets merged to master :stuck_out_tongue:

DanPurdy:
Nice catch, will sort that out.

bgriffith:
Looking good!

PR #81, opened 2015-09-02
Add Empty Args rule

### LINKED ISSUE #38: New Rule: Empty Args
With Sass, if a function or mixin doesn't have any arguments, the parenthesis can be left off of the definition and, in the case of mixins, the invocation as well

bgriffith:
This PR adds a rule for Empty Args which enforces whether or not parenthesis should be included if no arguments are defined or used, when declaring or invoking a mixin.

Closes #38 

### DIFF
--- a/docs/rules/empty-args.md
+++ b/docs/rules/empty-args.md
@@ -1,0 +1,33 @@
+# Empty Args
+
+Rule `empty-args` will enforce whether or not parenthesis should be included if no arguments are defined or used, when declaring or invoking a mixin.
+
+## Options
+
+* `include`: `true`/`false` (defaults to `false`)
+
+## Examples
+
+When `include: false`, the following are allowed. When `include: true`, the following are disallowed:
+
+```scss
+@mixin bar {
+  color: orange;
+}
+
+.bar {
+  @include bar;
+}
+```
+
+When `include: true`, the following are allowed. When `include: false`, the following are disallowed:
+
+```scss
+@mixin foo() {
+  color: red;
+}
+
+.foo {
+  @include foo();
+}
+```
--- a/docs/rules/empty-args.md
+++ b/docs/rules/empty-args.md
@@ -12,7 +12,7 @@ When `include: false`, the following are allowed. When `include: true`, the foll
 
 ```scss
 @mixin bar {
-  color: orange;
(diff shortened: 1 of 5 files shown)


Snugug:

bgriffith:
Fixed!

Snugug:
Doh!

bgriffith:
Fixed merge-conflict for now.

DanPurdy:
more of those wonderful conflicts!

Snugug:
Fo Real Though.

What the hell is going on w/all of these merge conflicts?

DanPurdy:
I think it's due to trying to keep all the tests and the yml file alphabetically organised, when you're working on two features at the same time though whichever gets merged first will cause conflicts in the next one..

I think we should perhaps split that main.js testing file out into seperate test files? What do you think, should mean the only conflicts then are the default lint.yml file.

Snugug:
Yah, I think that's a fine way to do it; split the tests for each rule out in to their own files.

We should (and it's mostly my fault cause this is how I started) test not just the defaults, but options as well.

If we split the tests for each rule out, we need to make sure they're all getting covered by the lint test. I recommend a `rules` folder that everything goes in to

DanPurdy:
Agreed, I've been adding tests for all the options to the rules I've touched but happy to go back and add more to the others..

docs/rules/empty-args.md
```diff
@@ -0,0 +1,33 @@
...
+
+* `include`: `true`/`false` (defaults to `false`)
+
+## Examples
+
+When `include: false`, the following are allowed. When `include: true`, the following are disallowed:
+
+```scss
+@mixin bar {
+  color: orange;
```
DanPurdy: We need to stop breaking our own linting rules in examples :smile:

docs/rules/empty-args.md
```diff
@@ -0,0 +1,33 @@
...
+.bar {
+  @include bar;
+}
+```
+
+When `include: true`, the following are allowed. When `include: false`, the following are disallowed:
+
+```scss
+@mixin foo() {
+  color: red;
```
DanPurdy: same as above

DanPurdy:
Awesome, looks good!

Issue #70, opened 2015-09-01
Feature Request: Enable Disabling Linters via Source


mherchel:
SCSS-lint has the ability to disable the linters within the source files

https://github.com/brigade/scss-lint#disabling-linters-via-source

This is unfortunately important for my project's workflow :)

Snugug:
I'm sorry to hear that :stuck_out_tongue_closed_eyes: 

This isn't a high priority at the moment with a slew of rules that need to be written and some bugs to be squashed, but I'd be happy to take a PR for it. I'd expect the abstraction for determining if a rule should be enabled/disabled to live in [helpers](https://github.com/sasstools/sass-lint/blob/master/lib/helpers.js)

cgtm:
Yeah, this would be hugely useful for the large project I'm working on. We've only recently started linting our scss files (I know!) and we've been using scss-lint, peppering all of our current files with overrides to get them to pass for now (as described by mherchel above).

As we're moving from scss-lint to sass-lint (ditching as much ruby stuff as possible), it would be great to have this functionality. Otherwise I can see needing to create multiple rulesets for different sets of files (being loose with everything already in production and tight with all new stuff).

I think there's going to be a bit move away from all the ruby based stuff (ruby-sass, compass, scss-lint, etc.), so this will be more and more of a 'requirement' for folks to ease the transition.

So, +1 from me!

benthemonkey:
This is definitely something that should be saved for when we majorly restructure the code to perform a single traversal of the AST (see #5)

ctumolosus:
Also, useful if you want people to use this package while `gonzales-pe` is trying to figure out how to solve #321 without having to ignore the whole file. I have the same issue using other tools that rely on it, but they provide a way to disable/enable sections of the source code. I've been waiting for interpolation to be fixed in `gonzales-pe` for a while.

DanPurdy:
I'm afraid this wouldn't solve that problem @ctumolosus. The issue is is that even with this feature `gonzales-pe` would still throw an error on interpolated code even if it was set to be ignored. We can't ignore code without first parsing it to know what it is. `gonzales-pe` parses the whole file and then we act upon what it returns. The way you are looking at this working to ignore interpolation would actually need to be a `gonzales-pe` feature to ignore code as it parses it.

natorojr:
FWIW, I'm another sass-lint user in desperate need of this feature.

Glad to see it's being worked on / considered.

mknadler:
[...]

PR #83, opened 2015-09-02
Add hex validation rule

### LINKED ISSUE #82: New Rule: Hex Validation
Rule hex-validation should enforce all hexadecimal values are of a valid format

as per [SCSS Lint](https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#hexvalidation)

DanPurdy:
Rule hex-validation will enforce all hexadecimal values are of a valid format.

* 3 or 6 characters long
* 0-9 a-f A-F

Adds a new `isValidHex` helper function

closes #82 

### DIFF
--- a/docs/rules/hex-validation.md
+++ b/docs/rules/hex-validation.md
@@ -1,0 +1,34 @@
+# Hex Validation
+
+Rule `hex-validation` will enforce the validity of hexadecimal values.
+
+## Examples
+
+When enabled any invalid hexadecimal characters will generate a warning/error:
+
+```scss
+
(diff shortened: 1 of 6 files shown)


Snugug:

DanPurdy:
Similar question to the other PR really. 

As some of these rules overlap we really want to be testing each rule in isolation, I'm thinking we disable any specifically conflicting linting rules on each test to ensure we aren't seeing false positives. I think something like #70 in the future will remove this need possibly.

Any thoughts? For now I'll push a change that specifically disables and enables rules on tests.

Issue #80, opened 2015-09-02
Invalid warning on Mixins should come before declarations when having multiple includes


jjanssen:
I am currently working with sass-lint 1.1.0

When using multiple mixins within a selector the rule for 'Mixins should come before declarations' starts throwing invalid warnings when using 3 or more. In another scenario the prepending selector can also cause this issue to throw a warning.

I have setup this little scss file to reproduce this issue:

```
@mixin foo($content) {
    content: $content;
}

@mixin bar {
    background-color: red;
}

@mixin hello {
    border-radius: 3px;
}


.beverages {

    .coffee {
        width: 100px;
        height: 100px;
    }    

    .tea {
        @include foo('yo');
        @include bar;
        @include hello;
        display: block;
        width: 100px;
        height: 100px;
    }
}
```

This snippet throws the following warning through the CLI
```
scss/mixin.scss
  24:9  warning  Mixins should come before declarations  mixins-before-declarations
```

This doesn't occur in the following cases:
1. When disabling the selector `.coffee` the warning is gone.
2. When disabling `@include hello;`. The other includes do not make a difference.

DanPurdy:
Thanks for this report @jjanssen it made tracking the issue a little easier. Hopefully fixed now, will just await review from the others.

jjanssen:
No problem! Seemed you guys put a lot of effort in this package the past few weeks. So if I can help by narrowing down the use case then it makes it easier for everybody :)

PR #88, opened 2015-09-03
Update mixin-before-declaration rule fix #80

DanPurdy:
The current mixin before declaration rule was acting strangely due to the way Gonzales is passing it's node counter value, I rewrote the rule to take into account a 'depth' value when traversing across embedded (can't think of the right word!) rules.

It first looks for include statements and then traverses only the parent element so it's a little easier to keep track of where we are. It then looks for a ruleset block in order to increment depth.

Also added some more test cases as brought up in #80 and added some tests for passing in values to the exclude property.

Hopefully this is ok to rewrite?

fixes #80 

### DIFF
--- a/lib/rules/mixins-before-declarations.js
+++ b/lib/rules/mixins-before-declarations.js
@@ -11,35 +11,37 @@ module.exports = {
     ]
   },
   'detect': function (ast, parser) {
-    var result = [],
-        error;
+    var result = [];
+    var error;
 
-    ast.traverseByType('block', function (block) {
-      var lastDeclaration = null;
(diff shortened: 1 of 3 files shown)


Snugug:
The one comment; otherwise GTG

DanPurdy:
Just a preference for me with 2 spaces, don't like dangling vars to 4 spaces. Happy to shift it over though!

bgriffith:
Let's keep a standard. We're using it throughout.

Snugug:
Agreed.

DanPurdy:
Agreed too, done!

PR #92, opened 2015-09-03
Change \n to os.EOL to work with Windows

### LINKED ISSUE #65: False positive for 'final-newline' rule on windows
The `final-newline` rule is always reporting a warning for me and I believe it's because of the way Windows reports newlines.

Because they are actually `\r\n`, [this line](https://github.com/sasstools/sass-lint/blob/master/lib/rules/final-newline.js#L27) (`last.content !== '\n'`) is always true.

If I test for `\r\n`as well the rule behaves as expected on Windows (although I don't have a mac handy to test it on).
```js
if (last.content !== '\n' && last.content !== '\r\n' && parser.options.include) {
```
Or a shorter version...
```js
if (!last.content.match(/\r?\n/) && parser.options.include) {
```

Snugug:
Resolves #65 

### DIFF
--- a/lib/rules/empty-line-between-blocks.js
+++ b/lib/rules/empty-line-between-blocks.js
@@ -1,6 +1,7 @@
 'use strict';
 
-var helpers = require('../helpers');
+var helpers = require('../helpers'),
+    os = require('os');
 
 var findNearestReturn = function (parent, i) {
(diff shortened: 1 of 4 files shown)


bgriffith:
Looking good!

PR #96, opened 2015-09-04
Add rule: Color variable

### LINKED ISSUE #27: New Rule: Color Variables
As per [SCSS Lint](https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#colorvariable)

DanPurdy:
Add rule color variable to restrict the use of color literals and color functions to variable declarations. This is to avoid the following:

```scss
.foo {
  color: blue;
  border-color: rgb(255, 255, 255);
  background-color: #fff
}
```

There are 5 possible warnings currently as the rule is fairly encompassing. Note that the values in quotes are dynamically produced from the actual linted code.

> color literals such as 'red' should only be used in variable declarations
 
> color functions such as 'rgb' should only be used in variable declarations
 
> color literals such as 'fff' should not be passed to functions, use variables
 
> color functions such as 'hsl' should not be passed to functions, use variables
 
> color literals such as 'red' should not be used as map identifiers

closes #27 

also note that duplicate-property rule was causing false positives in the tests therefore I've manually disabled it in each rule it was causing issues, will look to address this further soon.


DanPurdy:
Ready for review, especially interested to hear you guys' thoughts on the literals.yml file getting pulled in and whether we all agree this is the best way to keep this list maintainable?

Snugug:
One initial though; does `rgba($red, .5)` work with these turned on? I think that'd be something that we'd need to account for, at least in options, as it's a valid way to change opacity without changing the actual base color.

Snugug:
In that case, what about `mix` or `lighten` or `darken`? To me, the spirit of the Color Variables rule is that you don't use colors directly, not that you absolutely can't manipulate colors outside of a variable

Snugug:
I think a whitelist may not be smart enough, especially for `rgb/a` and `hsl/a`; those are both CSS ways of defining colors _and_ Sass color functions. I think there should either be a toggle to allow adding alphas to variable defined colors or not, or allowing variable usage in those.

To me, that looks like each color function in Sass only allowing variables as arguments except for possibly the alpha value where applicable.

DanPurdy:
OK so out of the disallowed functions the only conceivable one you can use with a single variable colour is `rgba` therefore to allow its use there is now a `allowRgba` flag which when set allows you to use `rgba` directly on any property but it MUST have a variable colour passed to it, it will fail if number style rgb components are passed.

Added another test to make sure of this option too.

The other SASS color functions such as `lighten` and `mix` are unaffected and can be passed color variables as before. Errors will still be generated if literals are used though.

This is quite a hefty rule and I'm thinking this may have gone over the top, perhaps break this down into seperate rules?

Snugug:
For consistency stake, options should be dasherized, not camelCase, so `allow-rgba` instead of `allowRgba`

DanPurdy:
My bad, knew something looked off, thanks for the catch :+1:

DanPurdy:
I remembered why it was camel case now.. you have to do the ugly `parser.options['allow-rgba'] in order to use hyphens in variable names.. Knew there was a reason..

Snugug:
This is true, but I'm more concerned with user experience where dasherize is better, so please change it

DanPurdy:
Definitely!

DanPurdy:
Oh and it was already changed when I wrote that!

Snugug:
The literals appear to be missing a couple of colors (literally, 2 colors, based on NodeList length).

Can you please make sure the literals length aligns with https://drafts.csswg.org/css-color/#named-colors (also [X11 Colors](https://en.wikipedia.org/wiki/X11_color_names).

What are our thoughts on `transparent` and `currentcolor` being in here? I think `currentcolor` is fine for people to use, and am leaning towards allowing `transparent` too, but would like thoughts.

DanPurdy:
https://drafts.csswg.org/css-color/#named-colors is exactly where I pulled the original list from and triple checked it, they're all there, I don't see any missing ones.. This branch originally had the literals in the commit but the other color rule I did for color keyword in #102 ended up being merged first with the more up to date list. 

As for `transparent` and `currentcolor` I think they are perfectly acceptable to still use and keep as they aren't actually denoting a physical color.

Snugug:
Maybe I just can't count :stuck_out_tongue_closed_eyes: 

I'm good with this

bgriffith:
I also think they should be exceptions :+1:

bgriffith:
Let's do it!

pamo:
Was this rule deprecated at any point? I no longer see it in the documentation but I'd like to use it.

nocarroll:
For anyone scratching their head on this as I was, the rule is now called `no-color-literals` instead of `color-variable`

PR #100, opened 2015-09-04
Add rule: Variable For Property

### LINKED ISSUE #33: New Rule: Variable for Property
As per [SCSS Lint](https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#variableforproperty)

DanPurdy:
Rule `variable-for-property` will enforce the use of variables as values for specified properties. Properties are passed into the properties array in the config file. By default no properties are specified.

The warning message encountered will be:

> Values for properties of type 'IDENTIFIER' may only be variables

closes #33 


bgriffith:
Just a thought with this one - would you likely want it to flag up issues in mixins? Or just within a selector. Maybe a flag?

DanPurdy:
I think it should be global for now but maybe we could open an issue to see if there's interest in that?

bgriffith:
Sure - happy with that approach!

DanPurdy:
Ready to go!

PR #102, opened 2015-09-04
Add rule: Color Keyword

### LINKED ISSUE #101: New Rule: Color Keywords
Rule to enforce the use of hexadecimal colour values rather than CSS literal ones.

as per [SCSS Lint](https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#colorkeyword)

DanPurdy:
Rule `color-keyword` enforces the use of hexadecimal colour values instead of CSS colour literals.

A warning message will display to tell you which colour literal needs replacing and it will also display the correct hexadecimal value.

> Color 'red' should be written in its hexadecimal form `#ff0000`

closes #101 


bgriffith:
Need to add `rebeccapurple` literal + hex to literals.yml

DanPurdy:
Any others? I scoured around for a definitive list..

bgriffith:
Only spotted that one! Though I don't know them all off by heart. The list is probably good, `rebeccapurple` is a pretty new one while the others haven't changed for a long time so that's probably the reason why.

DanPurdy:
Sorted, I think it is the only one missing I double checked all the usual resources..

bgriffith:
Looking good! :shipit:

PR #103, opened 2015-09-04
Feature/explicit rules

### LINKED ISSUE #58: Rules explicitly turned off if not listed ESLint style
It's not clear if rules *not* present in `sass-lint.yml` files are turned off automatically (ESLint 1.0 style) or not. 

With early runs of `gulp-sass-lint`, that does not seem to be the case; for codebases that have both `Sass` & `SCSS` litn files, it seems there's issues propagating from `gonzales-pe` unless `trailing-semicolon` is explicitly set to `0` rather than merely omitting the rule altogether:

screen shot 2015-08-30 at 10 15 14 pm

Seemingly any rule not listing in the `sass-lint.yml` should be turned off (severity of `0`)  since that's the clearest way for users of the plugin-in to communicate they are not concerned of that rule being applied to their codebase.

Overtime, it can be more work for them to apply rules they don't care about in their codebase than the rules they care about as rules become simultaneously more useful & complex.

Snugug:
Resolves #58

### DIFF
--- a/docs/options/merge-default-rules.md
+++ b/docs/options/merge-default-rules.md
@@ -1,0 +1,9 @@
+# Merge Default Rules
+
+Option `merge-default-rules` will determine whether the rules present in a user's external configuration file or passed directly in to Sass Lint will override or merge with the default rules present in Sass Lint. All other configuration options will still be merged.
+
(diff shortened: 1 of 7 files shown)


DanPurdy:
Looks good to me.

Nice! :+1:

bgriffith:
Having just tried this out it makes testing so more awesome!

Issue #98, opened 2015-09-04
Space After Colons doesn't work inside of arguments


Snugug:
For instance, inside of Media Query arguments:

https://github.com/bgriffith/sass-lint/commit/2ceb329be1e014ddf7c71508eabf37d59a223e07#diff-9749c71a6762edb9480fb15c79950231R23

bgriffith:
Had a quick look and the issue is that properties inside a media query aren't classified as a `propertyDelimiter` so he doesn't include them in the loop.

DanPurdy:
What are they classed as?

bgriffith:
Just operators

```
{
                      "type": "operator",
                      "content": ":",
                      "syntax": "scss",
                      "start": {
                        "line": 2,
                        "column": 20
                      },
                      "end": {
                        "line": 2,
                        "column": 20
                      }
                    }
```

bgriffith:
Shouldn't the title of this be 'Space after colon..? instead of parens?

Snugug:
Yah… probably…

PR #106, opened 2015-09-05
Fix Space Before Colon and paren issue

bgriffith:
Fixes an issue with Space Before Colon that was fixed for the Space After Colon rule. The issue is that colons within parens are not linted.

See issue #98 


DanPurdy:
Looking good

PR #107, opened 2015-09-05
Feature/write output to file

### LINKED ISSUE #48: Write results to file
One of the features of SCSS Lint that I found useful was the ability to have the results written to a file, with the ability to specify the file path. This would be a really useful addition.

I think it would be useful to silence the console output if writing to a file is chosen.

Snugug:
Resolves #48

### DIFF
--- a/docs/options/formatter.md
+++ b/docs/options/formatter.md
@@ -1,0 +1,7 @@
+# Formatter
+
+Option `formatter` will determine how results from Sass Lint will be formatted for output.
+
+## Options
+
+Any of the supported [ESLint Formatters](https://github.com/eslint/eslint/tree/master/lib/formatters) can be used. Defaults to `stylish`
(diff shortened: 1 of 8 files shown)


DanPurdy:
Is there any plans to add defining an output directory?

Snugug:
I had only considered outputting results to a file, not a way of outputting multiple files to a directory

DanPurdy:
Makes sense, this is great though. Eslint formatters will be a massive plus to people. 

Great stuff, @bgriffith you should give this a once over too

Snugug:
Haha, the use of ESLint formatters was mostly because I'm too lazy to write my own (and for the life of me I can't figure out why they haven't been blown out in to their own packages yet)

DanPurdy:
Ha yeah, if they aint broken why fix and all that...

bgriffith:
Looking good.

PR #113, opened 2015-09-05
Enhance Property Sort Order with presets

### LINKED ISSUE #20: Enhance property-sort-order (other preset orders apart from alphabetical)
One for a rainy day.. It would be great if we could have a few preset orders to chose from (biased for Recess) when configuring the property-sort-order, similar to scss-lint (https://github.com/brigade/scss-lint/tree/master/data/property-sort-orders) and the config (https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#propertysortorder)

bgriffith:
This PR enhances `property sort order` rule with the addition of presets - RECESS, SMACSS and Concentric. It also adds a few helper functions to allow easy loading of config files, retrieval of file extensions and filenames.

Tests for presets have also been included.

Closes #20 


DanPurdy:
This is good stuff!

Seen as you have the capability to load config files already I think a great addition would be to accept custom order `.yml` files rather than just passing an array of properties in `sass-lint.yml`. What do you think about that?

Once this PR is merged in I'll move ahead and update the colour literal loading into these new helper functions..

Snugug:
Some comments before it gets merged in.

bgriffith:
Updated to use path functions and moved the config files to the config dir.

Snugug:
In the future, instead of one commit to add the files and one commit to delete, them, just include a single rename commit.

A couple more comments, but otherwise looks good!

Snugug:
Please also add test for the new helper function to our list of `helper` tests

bgriffith:
Will do.

Snugug:
Thanks!

Snugug:
I'm confused by the last commits; they seem to be reverting to manual filename and extension parsing instead of using `path`

bgriffith:
Just updated - sorry for the confusion.

Snugug:
Great, thanks.

PR #115, opened 2015-09-05
Fix CLI config when passing files as arguments

### LINKED ISSUE #114: CLI --config is ignored
When running `sass-lint --config .my-scss-lint.yml index.scss` using the CLI, the `--config` parameter is ignored. The bug is, that `configPath` is not passed to `sassLint.lintText` in `index.js`.

skovhus:
Resolves #114 

DCO 1.1 Signed-off-by: Kenneth Skovhus <kenneth.skovhus@gmail.com>

### DIFF
--- a/index.js
+++ b/index.js
@@ -74,7 +74,7 @@ sassLint.lintFiles = function (files, options, configPath) {
       'text': fs.readFileSync(file),
       'format': path.extname(file).replace('.', ''),
       'filename': file
-    }, options);
+    }, options, configPath);
     results.push(lint);
   });
 


DanPurdy:
Thanks for this @skovhus.

Confirmed and tested this as fixing the issue.

skovhus:
@DanPurdy thanks. Any ETA of when this will be released? I would like to use this for a sublimelinter plugin for sass-lint.

DanPurdy:
@skovhus it should be very soon, we're tracking one more issue for [1.2.0](https://github.com/sasstools/sass-lint/milestones/1.2.0) which is currently under review. Don't take my word for it but I would imagine it will be released sometime this week. There have been a lot of additions in this new version so we are keen to get it out.

Issue #116, opened 2015-09-06
Helper function unit tests


DanPurdy:
Currently there are no unit tests for the helper methods.

#### Todo

- ~~log~~
- [x] propertySearch
- [x] isEqual
- [x] isUnique
- [x] addUnique
- [x] sortDetects
- [x] isNumber
- [x] isUpperCase
- [x] isLowerCase
- [x] isValidHex

DanPurdy:
Added isEqual tests in 792fb50

DanPurdy:
Closed for now with #117

PR #117, opened 2015-09-06
Add tests for helper methods

DanPurdy:
Add tests for helper methods

Opening this PR early to track progress on these tests. Will update as they happen.

References  #116

### DIFF
--- a/tests/helpers.js
+++ b/tests/helpers.js
@@ -3,11 +3,23 @@
 var assert = require('assert'),
     helpers = require('../lib/helpers');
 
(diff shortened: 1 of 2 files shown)


DanPurdy:
Ready for review. Added tests for all helper functions excluding log.. Mainly because that is just a console.log and we don't use it anywhere in the code, only in dev/debugging..

Snugug:
Yah, I've noticed sometimes ESLint's indentation freaks out and is weird for some reason… Don't know why

Issue #91, opened 2015-09-03
Empty Scss file breaks linting


elidupuis:
I'm running 1.1.0 CLI and getting the following error:
```
/Users/eli/.nvm/v0.12.7/lib/node_modules/sass-lint/lib/rules/extends-before-mixins.js:11
    ast.traverseByType('block', function (block) {
        ^
TypeError: undefined is not a function
    at Object.module.exports.detect (/Users/eli/.nvm/v0.12.7/lib/node_modules/sass-lint/lib/rules/extends-before-mixins.js:11:9)
    at /Users/eli/.nvm/v0.12.7/lib/node_modules/sass-lint/index.js:30:25
    at Array.forEach (native)
    at Function.sassLint.lintText (/Users/eli/.nvm/v0.12.7/lib/node_modules/sass-lint/index.js:29:9)
    at /Users/eli/.nvm/v0.12.7/lib/node_modules/sass-lint/index.js:73:21
    at Array.forEach (native)
    at Function.sassLint.lintFiles (/Users/eli/.nvm/v0.12.7/lib/node_modules/sass-lint/index.js:72:9)
    at detectPattern (/Users/eli/.nvm/v0.12.7/lib/node_modules/sass-lint/bin/sass-lint.js:16:18)
    at /Users/eli/.nvm/v0.12.7/lib/node_modules/sass-lint/bin/sass-lint.js:58:5
    at Array.forEach (native)
```

I was able to track this down to an empty `.scss` file. After deleting the empty file, everything runs fine!

PR #120, opened 2015-09-06
Add Test for Empty File

Snugug:
Only run parsing for files with stuff in them

Resolves sasstools/sass-lint#91


DanPurdy:
Nice one :+1:

elidupuis:
:+1:

PR #124, opened 2015-09-06
New rule: space around operator

### LINKED ISSUE #31: New Rule: Space Around Operator
As per [SCSS Lint](https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#spacearoundoperator)

skovhus:
This resolves #31. 

Due to bugs in `gonzales-pe` the following will currently fail:
```sass
body {
  background: url('a'  +  'b.png');  // will not throw an error
}

$foo: 90px   %10px;  // will crash
```

DCO 1.1 Signed-off-by: Kenneth Skovhus kenneth.skovhus@gmail.com

### DIFF
--- a/lib/rules/space-around-operator.js
+++ b/lib/rules/space-around-operator.js
@@ -1,0 +1,63 @@
+'use strict';
+
+var helpers = require('../helpers');
+
+var validateExpression = function (parser, result, expressionNode) {
+  if (!expressionNode) {
+    return;
+  }
+
+  for (var j = 0; j < expressionNode.content.length; j += 2) {
+    // every second node should be a single space
+    var node = expressionNode.content[j + 1];
+    if (node && !node.is('space')) {
+      helpers.addUnique(result, {
+        'ruleId': parser.rule.name,
+        'line': node.start.line,
+        'column': node.start.column,
+        'message': 'Value should be followed by a space',
+        'severity': parser.severity
+      });
+      break;
+    }
+    else if (node && node.is('space') && node.content !== ' ') {
+      helpers.addUnique(result, {
+        'ruleId': parser.rule.name,
+        'line': node.start.line,
+        'column': node.start.column,
+        'message': 'Value should be followed by a single space',
+        'severity': parser.severity
+      });
+      break;
+    }
+  }
+};
+
+module.exports = {
+  'name': 'space-around-operator',
+  'defaults': {
+    'include': true
+  },
+  'detect': function (ast, parser) {
+    var result = [];
+
+    ast.traverseByType('atruleb', function (exp) {
+      exp.content.forEach(function (node) {
+        if (node.is('parentheses')) {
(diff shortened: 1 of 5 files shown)


lib/rules/space-around-operator.js
```diff
@@ -0,0 +1,63 @@
...
+        'message': 'Value should be followed by a single space',
+        'severity': parser.severity
+      });
+      break;
+    }
+  }
+};
+
+module.exports = {
+  'name': 'space-around-colon',
```
bgriffith: Should be `space-around-operator`

skovhus: @bgriffith Good catch. Should be fixed now.

DanPurdy:
Hey @skovhus thanks for this, we'll leave it on review until 1.2.0 is out. I believe we are planning to move to using a develop branch post 1.2 to enable us to simultaneously release hot fixes to master if necessary while continuing to develop and merge rules and features. 

As you pointed out too we're awaiting a new release of `gonzales-pe` too to fix some of the issues we've encountered.

skovhus:
@DanPurdy makes sense. Looking forward to 1.2.0. After that https://github.com/skovhus/SublimeLinter-contrib-sass-lint should be ready for use. : )

Out of curiosity, what were the considerations in choosing `gonzales-pe`? I haven't looked in to this, but I for me it would make sense to use the same parser/AST as `node-sass`. That should be pretty battle tested by now.

DanPurdy:
Yeah it's gonna be good to get 1.2.0 out, and a sublime linter plugin ready to go is also great news! :+1: 

I've only come on board recently as I'm looking forward to moving away from the ruby dependencies in some of our projects. @Snugug has talked about it in a few of the earlier issues though., but he will be able to tell you why.

Snugug:
Gonzales was chosen because Libsass (the compiler that Node Sass uses) doesn't provide its AST publicly and, in its current state, wouldn't be useable enough for linting.

There is an issue in Libsass to [expose the AST](https://github.com/sass/libsass/issues/1065), but until that lands (and the AST is robust enough to support writing a linter against), we can't use it for Sass Lint.

skovhus:
@Snugug Thanks for clarifying.

Another thing now that we are talking about the background of this project: what are your vision in terms of rules porting? Do you imagine sass-lint to port all scss-lint rules and their many options?

Snugug:
If you want to discuss this, please open an issue to discuss roadmap

Snugug:
This will need to be re-rolled off of the `develop` branch

DanPurdy:
Hey @skovhus this is ready for a new PR to be created in to develop if you have the time.

Thanks!

DanPurdy:
I'm going to close this for now. I've just run through it to update with sass tests and even a few of the scss tests aren't doing what they are supposed to. Will definitely need more work.

Issue #123, opened 2015-09-06
1.2 Cleanup


Snugug:
Go through all of everything and make sure everything is clean and consistent.

- [x] Options in rules documentation all formatted the same
- [x] Rules with no options all are in form of `no-{{rule}}`
- [x] All rule options are dasherized
- [x] All rules have documentation
- [x] Divide tests in to individual files

DanPurdy:
Could this include alphabetically organising [main.js](https://github.com/sasstools/sass-lint/blob/master/tests/main.js) so many merge conflicts from that file this time round. :smile:

Snugug:
I'll do you one better; divide each rule in to its own test

DanPurdy:
:heart: 

I was planning on it!

Snugug:
Tracking on https://github.com/Snugug/sass-lint/tree/cleanup/1.2

DanPurdy:
We've got fixes for #125 and #126 ready to go too. Do you want to roll these into 1.2.0 or get a 1.2.1 out straight after?

Snugug:
This rules test rewrite is a surprisingly big job, but on the bright side, I think I can drastically speed up our tests by doing so, in addition to making them easier to maintain

Snugug:
@DanPurdy I think we might as well roll them in to 1.2 as it's not released yet

Snugug:
@DanPurdy @bgriffith What are your thoughts on renaming `color-variable` to `no-color-literals`? The rule, like our other `no` rules, specifically disallows a single thing from happening and you cannot toggle it on or of, activating it toggles it on and that's it.

bgriffith:
If we're only enforcing not using it then I agree with using the `no` prefix. And since `no-color-variable` isn't what we're enforcing `no-color-literals` makes sense to me.

We should also be switching `color-keyword` to `no-color-keyword`.

Snugug:
I believe I already did color keyword

DanPurdy:
I think it makes sense, have you got any thoughts on #126 @Snugug @bgriffith I'm starting to think the extra option for preventing variables being named the same as colour keywords makes a lot of sense..

In the case you mention above with the `no-` prefix it would essentially mean a new rule which I actually think is cleaner IF you think this is necessary as a rule at all..

Snugug:
[...]

PR #130, opened 2015-09-07
Cleanup/1.2

Snugug:
Resolves https://github.com/sasstools/sass-lint/issues/123


bgriffith:
This is great! Should make maintaining the tests a lot easier! 

I've looked through the changes - all good!

lib/config/sass-lint.yml
```diff
@@ -18,23 +18,23 @@ rules:
...
-  no-duplicate-property: 1
+  no-duplicate-properties: 1
   no-empty-rulesets: 1
   no-extends: 0
   no-ids: 1
   no-important: 1
   no-warn: 1
+  no-color-keywords: 1
+  no-invalid-hex: 1
+  no-css-comments: 1
```
bgriffith: Do we want these to be ordered alphabetically?

Snugug: As these are our default values and we have a separate [sample config](https://github.com/sasstools/sass-lint/blob/master/docs/sass-lint.yml), I don't really care if it's alphabetical or not.

bgriffith: Yeah ok - fair point hah

PR #132, opened 2015-09-07
Fix no-color-keyword bug

### LINKED ISSUE #126: Bug: color keyword errors on variables named as literals
Using 1.1.0, when `color-keywords` is linted, the following is disallowed:

```
$blue: #0050a0;
$brand-primary: $blue;
```

DanPurdy:
Fixes the issue #126 

* Check for parent type `variable`
* Add example from #126 to test file

### DIFF
--- a/lib/rules/no-color-keywords.js
+++ b/lib/rules/no-color-keywords.js
@@ -16,7 +16,7 @@ module.exports = {
     ast.traverseByType('value', function (node) {
       node.traverse(function (elem, i, parent) {
         var isVariable = false;
-        
+
         if (parent) {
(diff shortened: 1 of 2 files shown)


Snugug:
:+1:

pzi:
I am wondering if this should be extended to `@function`s too?

My use case (that got highlighted as warning):

```sass
// _functions.sass
@function black($opacity)
  @return rgba(0, 0, 0, $opacity)

@function white($opacity)
  @return rgba(255, 255, 255, $opacity)

// _component.sass
.component
  box-shadow: 0 1px 1px black(0.3)
  background-color: white(0.8)
```

PR #145, opened 2015-09-09
Add unit test for ignored files

mven:
It seems there must be no space between each ignored pattern for glob to recognize them all. I fixed that, as well as added a unit test to test against ignored files. I ran `npm test` and it passed successfully.

```bash
cli
    ✓ should return help instructions (120ms)
    ✓ should return a version (119ms)
    ✓ should not include ignored paths (461ms)
```
...
```bash
 zero unit
    ✓ [include: false]
    ✓ [include: true]

  60 passing (219ms)
```

### DIFF
--- a/bin/sass-lint.js
+++ b/bin/sass-lint.js
@@ -31,7 +31,7 @@ program
   .version(meta.version)
   .usage('[options] <pattern>')
   .option('-c, --config [path]', 'path to custom config file')
-  .option('-i, --ignore [pattern]', 'pattern to ignore. For multiple ignores, separate each pattern by `, `')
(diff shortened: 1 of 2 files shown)


mven:
So close! lol.

mven:
Looks like the the error is that sass-lint couldn't be installed globally on node 4.0 
https://travis-ci.org/sasstools/sass-lint/jobs/79392140#L136

DanPurdy:
@Snugug @mven whats the status here? Have we decided on a way forward?

DanPurdy:
Incoming PR to sort this

PR #136, opened 2015-09-08
Add Url Format rule

bgriffith:
Adds the `url-format` rule that disallows protocols and/or domains in urls.

When enabled, if a protocol or domain is found, the following warning message gets displayed:
> Protocols and domains in URLs are disallowed

DCO 1.1 Signed-off-by: Ben Griffith gt11687@gmail.com

Closes #94

PR #147, opened 2015-09-09
Rename Url Format to No Url Protocols

bgriffith:
Rename URL Format to No URL Protocols as noted in #136 


Snugug:
Perf. Thanks!

PR #156, opened 2015-09-11
Enhance no-duplicate-properties rule to allow exceptions

### LINKED ISSUE #142: @font-face multiple src-definition throws unwanted duplicate property warning
Since 1.2.0 the duplicate property rule got introduced. When using the `@font-face` declaration you it requires use of multiple `src` declarations if you want to support version prior to IE9. Like in the following example:

```
@font-face {
    font-family: 'custom-font';
    src: url('custom-font.eot?3q6lb2');
    src: url('custom-font.eot?#iefix3q6lb2') format('embedded-opentype'),
    url('custom-font.ttf') format('truetype'),
    url('custom-font.woff') format('woff'),
    url('custom-font.svg?#custom-font') format('svg');
    font-weight: normal;
    font-style: normal;
}
```

The output from this is:
```4:5  warning  Duplicate properties are not allowed within a block  no-duplicate-properties```

I am not sure whether to completely file this as a bug. In the future this could be fixed by disabling linter through via source (#70). 

On the other hand, when validating for a `DuplicateProperty` through the Ruby SCSS Linter it validates without any warnings.

bgriffith:
Adds a feature to exclude properties from the no-duplicate-properties rule. Closes #142 

### DIFF
--- a/docs/rules/no-duplicate-properties.md
+++ b/docs/rules/no-duplicate-properties.md
@@ -2,6 +2,11 @@
 
 Rule `no-duplicate-properties` will enforce that duplicate properties are not allowed within the same block.
 
+## Options
+
+* `exclude`: `[array of property names to be excluded from this rule]` (defaults to empty array `[]`)
+
+
 ## Examples
 
 When enabled, the following are disallowed:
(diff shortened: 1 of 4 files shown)


bgriffith:
Todo: Still need to add new feature to docs :memo:

DanPurdy:
Awesome, this looks good. Makes sense to have this option too

PR #158, opened 2015-09-12
Hotfix master - Fix custom options overwriting defaults between assertions

### LINKED ISSUE #157: Custom rule options overwriting defaults between test assertions
Currently custom rule options are overwriting the defaults for any assertions that follow.

See example below.

The defaults:
```js
...
'defaults': {
    'allow-element-with-attribute': false,
    'allow-element-with-class': false,
    'allow-element-with-id': false
  },
...
```

First assertion will correctly allow elements with attributes.
```js
  it('[attribute: true, class: false, id: false]', function (done) {
    lint.test(file, {
      'no-qualifying-elements': [
        1,
        {
          'allow-element-with-attribute': true
        }
      ]
    }, function (data) {
      lint.assert.equal(2, data.warningCount);
      done();
    });
  });
```

Second assert takes defaults to be
```js
...
'defaults': {
    'allow-element-with-attribute': true,
    'allow-element-with-class': false,
    'allow-element-with-id': false
  },
...
```

Therefore second assertion will incorrectly allow elements for both attribute and class
```js
  it('[attribute: false, class: true, id: false]', function (done) {
    lint.test(file, {
      'no-qualifying-elements': [
        1,
        {
          'allow-element-with-class': true
        }
      ]
    }, function (data) {
      lint.assert.equal(2, data.warningCount);
      done();
    });
  });
```

bgriffith:
This fixes an issue where custom rule options would overwrite the defaults for any other assertions that followed.

This was done to the `merge.recursive` function not cloning before merging, therefore overwriting the original.

Closes #157 


DanPurdy:
:+1:

PR #161, opened 2015-09-12
Add custom option tests to multiple rules

bgriffith:
Adds more tests to cover custom options in the existing rules.


DanPurdy:
:+1: more tests ftw!

Issue #15, opened 2015-08-27
New Rule: Shorthand


Snugug:
Add test if a property's value should be written as shorthand

DanPurdy:
Made up of 2 rules

* Enforce using the fewest values possible

```scss
margin: 0 2rem 0 2rem

//to

margin: 0 2rem;
```

* Enforce using a shorthand property / value where appropriate

```scss
margin-top : 0;
margin-right: 2rem;
margin-bottom: 0;
margin-left: 2rem;

//to

margin: 0 2rem;
```

## Todo
- [ ] Rule 1 - Enforce shorthand values
- [ ] Rule 2 - Enforce shorthand values and properties

DanPurdy:
Will add the second one here to a new rule

PR #167, opened 2015-09-13
Add shorthand values rule

DanPurdy:
Rule `shorthand-values` will enforce that values in their shorthand form are as concise as specified.

Related to #15 

Warning message
> Property `margin` should be written more concisely as `1px` instead of `1px 1px 1px 1px`

Don't include trailing zeros on numbers


DanPurdy:
Just need to know what you're doing with the Sass tests here @bgriffith before this should be merged, I'll add them in before we go.

bgriffith:
Good to see plenty of tests! Good to go :+1:

PR #168, opened 2015-09-14
Hotfix master - Fix issue with Single Line Per Selector and Sass syntax 

### LINKED ISSUE #165: TypeError: Cannot read property 'line' of undefined
When running `gulp-sass-lint` on my files I get the following error which seems to happen on the last rule. For exemple if I disable the `single-line-per-selector` rule the error happen on a different rule.

    /project_dir/node_modules/sass-lint/lib/rules/single-line-per-selector.js:20
                  'line': next.start.line,
                                    ^
    TypeError: Cannot read property 'line' of undefined
        at /project_dir/node_modules/sass-lint/lib/rules/single-line-per-selector.js:20:33
        at Node.forEach (/project_dir/node_modules/gonzales-pe/lib/node/basic-node.js:1:2053)
        at /project_dir/node_modules/sass-lint/lib/rules/single-line-per-selector.js:13:16
        at RootNode.traverseByType (/project_dir/node_modules/gonzales-pe/lib/node/root-node.js:1:1739)
        ...

    # node --version
    v0.12.5

    # osx 10.10.5

The error is gone if I change the following line: https://github.com/sasstools/sass-lint/blob/develop/lib/rules/single-line-per-selector.js#L16

    if (next && next.start) {

For reference here is a minimal file which cause the error: http://cl.ly/code/1b1o1U1y2q2n
And here the sass-lint.yml: http://cl.ly/code/39022c373X3q

bgriffith:
This PR fixes #165 where Single Line Per Selector rule would fail due to the parser interpreting the stylesheet differently to scss. 

### DIFF
--- a/lib/rules/single-line-per-selector.js
+++ b/lib/rules/single-line-per-selector.js
@@ -13,18 +13,20 @@ module.exports = {
       selector.forEach('delimiter', function (delimiter, i) {
         var next = selector.content[i + 1];
 
-        if (next.is('simpleSelector')) {
-          next = next.content[0];
-        }
+        if (next) {
+          if (next.is('simpleSelector')) {
+            next = next.content[0];
+          }
 
-        if (!(next.is('space') && next.content.indexOf(os.EOL) !== -1)) {
-          result = helpers.addUnique(result, {
-            'ruleId': parser.rule.name,
-            'line': next.start.line,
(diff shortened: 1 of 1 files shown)


DanPurdy:
Looks good here. :+1:

PR #174, opened 2015-09-15
Hotfix master - Fix No Extends to work on Sass Syntax files

### LINKED ISSUE #173: No extends not working with Sass Syntax
No Extends rule is currently not working with Sass syntax. The current method to check for nodes with the Extend type returns blank when using the Sass syntax.

bgriffith:
This PR fixes an issue where Extends were not being caught when looking for nodes of type extend. Turns out when using Sass syntax, the parser doesn't assign an extend type so I've implemented another way. This method is in line with the way we detect warns and debugs.

Closes #173 

### DIFF
--- a/lib/rules/no-extends.js
+++ b/lib/rules/no-extends.js
@@ -8,13 +8,17 @@ module.exports = {
   'detect': function (ast, parser) {
     var result = [];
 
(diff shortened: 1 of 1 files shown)


DanPurdy:
:+1:

PR #177, opened 2015-09-16
Change travis config to fix build errors

mven:
This should fix the build errors we were experiencing in Travis CI. Using `npm link` instead of installing from NPM. 

Referencing: #172 

DCO 1.1 Signed-off-by: Michael Vendivel vendivel@gmail.com

Issue #172, opened 2015-09-15
Testing environment for CLI


DanPurdy:
It's become clear that our Travis setup isn't currently adequate for testing our CLI or more specifically the unreleased work we do on it.

The problem lies in the fact that Travis when provisioning is installing `sass-lint` globally for CLI testing purposes but this obviously means it's only installing it at the last released version which at the time of writing is `1.2.0`.

Some of the work I've done for `1.2.1` #170 and `1.3.1` #152 is currently blocked with tests failing on Travis. Locally they are working fine but this is just because I've symlinked my globally installed sass-lint to my development copy.

This doesn't affect testing of non CLI work at all but is leading to quite a few false positives in the CLI tests.

Any thoughts on how to progress or how we'd like to modify our Travis setup.

mven:
As a temporary fix, do you think something like the following would work on Travis CI? It looks like it's working locally for me:

`var command = path.resolve(__dirname, '../bin/sass-lint.js') + ' -h';`

DanPurdy:
I'm open to all suggestions, personally I'd rather it wasn't code we add directly to `sass-lint` and we just get some way of making sure the CLI global install is up to date with development for testing, with pluggable rules on the cards I can see some fairly big core additions and CLI changes coming.

Definitely need to get this sorted soon though, `1.2.1` has a lot of very useful hotfixes in that we need to get out.

DanPurdy:
Nice one @mven

PR #178, opened 2015-09-16
Change travis config to fix build errors (for develop branch)

mven:
This should fix the build errors we were experiencing in Travis CI. Using npm link instead of installing from NPM.

Referencing: #172, #177 

DCO 1.1 Signed-off-by: Michael Vendivel vendivel@gmail.com


DanPurdy:
@mven ok if you update this branch with upstream/develop now we should be good to go

bgriffith:
Looking good!

PR #183, opened 2015-09-18
Fix issue of non-reseting prefixes

### LINKED ISSUE #182: Prefixes array being passed between assertions (no-vendor-prefixes)
Currently the `prefixes` array in the `no-vendor-prefixes` is not getting reset between assertions. This means that if you modify the array through adding or removing prefixes in an assertion, these changes will still exist for the next assertion.

Was this intended @DanPurdy  Only reason I ask is that the expected number of warnings is only true when this behaviour is allowed.

Easiest way to simulate is to create a new assertion with only the defaults.

SassLint: 1.2.0

bgriffith:
Fixes an issue where any modification to the prefixes array when testing using assertions gets carried across to the next. The issue is fixed by always setting the initial prefixes each time an assertion is run.

Closes #182 


DanPurdy:
I'm not sure about this.. We are adding extra steps to the linter to fix issues with our tests, this seems completely backwards to me, we should alter the way tests are set and run rather than modify the listing functionality?

bgriffith:
This only resets the prefixes for the rule. Assertions shouldn't be modifying code for the next one. Each assertion should have equal conditions?

I noticed this this issue when I create another assertion using the defaults and received a different amount of warnings compared to the first time I'd run it.

DanPurdy:
As we just discussed, I don't think the prefixes should be set within the module itself, though having the default prefixes as immutable within is good and a copy of this array should be the one we mutate. This will prevent the problem in the tests and also be just general good practice..

bgriffith:
Yep agreed! Update pushed.

bgriffith:
Going to tighten this up further.

DanPurdy:
Yeah this makes sense. :+1:

PR #187, opened 2015-09-19
Feature/reset defaults on new assertion color literals

### LINKED ISSUE #186: no-color-literals rule mutating a value - effecting assertions that follow
Similar issue to #182. `no-color-literals` is mutating a value (in this case the colorFunctions array) which is resulting in it being different for each assertion.

Sass-lint: 1.2.0

bgriffith:
Fixes an issue where any modification to the colorFunctions array when testing using assertions gets carried across to the next. The issue is fixed by always creating a copy of initial array and then modifying that instead when each time an assertion is run.

Closes #186


DanPurdy:
Should avoid a few testing problems :+1:

PR #189, opened 2015-09-19
Hotfix master - with extend rules sass syntax

### LINKED ISSUE #188: extends-before-declarations rule not working with Sass syntax
Currently the `extends-before-declarations` rule doesn't work with the Sass syntax. This is down to the parser interpreting the nodes differently.

This is related to #173 

SassLint: 1.2.0

bgriffith:
This PR fixes an issue where the `extends-before-declarations` rule doesn't work with the Sass syntax due to the way the parser interprets the syntax differently.

Related to issue #173 

Closes #188 


DanPurdy:
:+1: Looks good

PR #194, opened 2015-09-20
Add rule: No Misspelled Properties

### LINKED ISSUE #16: New Rule: Property Spelling
Add Rule to check to see if properties are spelled correctly to help avoid errors

DanPurdy:
Rule `no-misspelled-properties` will enforce the correct spelling of CSS properties and prevent the use of unknown CSS properties.

A warning
> 'Property `property-name` appears to be spelt incorrectly'

will be shown if the property doesn't appear in the master properties list or within the user specified `extra-properties` list.

The master property list was derived from the following sources:
* [w3.org - 1](http://www.w3.org/TR/CSS21/propidx.html) & [w3.org - 2](http://www.w3.org/Style/CSS/all-properties)
* [MDN - CSS Reference](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference)
* [meiert.com](http://meiert.com/en/indices/css-properties/)
* [scss-lint](https://github.com/brigade/scss-lint/blob/master/data/properties.txt)

closes #16 


lib/config/sass-lint.yml
```diff
@@ -33,6 +33,9 @@ rules:
   no-vendor-prefixes: 1
   no-warn: 1
 
+  # Helpers
```
DanPurdy: If anyone has suggestions about what group this should be in I'm all ears.

joshuacc:
Not a big deal, but "spelt" confused this American for a moment. I didn't even realize that spelling was still current outside the U.S.

DanPurdy:
Blimey guvna, it never even crossed my mind, made a bit of a dog's dinner of that! :wink: 

Would it be 'spelled' for our friends across the pond? It would probably make sense to stick to U.S. 'english'

lib/config/sass-lint.yml
Snugug: Potential Errors I think would be better than Helpers

DanPurdy: Ok will change to that, makes more descriptive sense. Thanks!

DanPurdy:
Have updated to use 'spelled' instead of 'spelt', looks so strange and wrong to me now! :smile: But let's leave it like that, I'll add it to my list of things to remember along with 'color' 'colour' and 'aluminum' 'aluminium'

joshuacc:
Apparently the U.S. standardized on "spelled" about a hundred years ago. Stood out to me immediately as old-fashioned. Of course, you could always rename the project to "Ye Olde Sass Linterre." :grin:

bgriffith:
Haha I'd better check this one over and merge quick :grin: There be a storm approachin'.

One quick comment; since this is a disallow (as you wouldn't even want to force incorrect spelling) i'd change the name to something like `no-incorrect-property-spelling` or a similar shorter version

Snugug:
How about no-misspelled-properties

bgriffith:
Yeah sounds good :+1:

DanPurdy:
Will update later on :+1:

DanPurdy:
Updated to `no-misspelled-properties`

bgriffith:
Didn't know half those properties existed. Looking good.

PR #196, opened 2015-09-20
Add rule: No transition all

### LINKED ISSUE #195: New Rule: No Transition All
As per [scss lint](https://github.com/brigade/scss-lint/tree/master/lib/scss_lint/linter#transitionall)

Prevent the use of the all keyword within the `transition` and `transition-property` properties.

DanPurdy:
Rule `no-transition-all` will enforce whether the keyword `all` can be used with the `transition` property.

The warning message below will be displayed if the keyword `all` is used as a value for `transition` or `transition-property`.

> The keyword `all` should not be used with the property `property-name`

closes #195 


bgriffith:
Update this and I'll merge :+1:

PR #205, opened 2015-09-21
Hotfix master - Fix Sass syntax issue with clean-import-paths rule

### LINKED ISSUE #179: Sass Syntax - Clean import paths issues
There are issues with the `clean-import-paths` rule when using the sass syntax as it's valid syntax to use without quotes. See guide http://sass-lang.com/guide

This in turn breaks the way we're detecting strings and validating them so a new way will need to be thought up.

bgriffith:
clean-import-paths doesn't work with Sass syntax. This fixes that issue.

Fixes #179 

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -149,4 +149,8 @@ helpers.loadConfigFile = function (configPath) {
   return file;
 };
 
+helpers.stripQuotes = function (str) {
+  return str.substring(1, str.length - 1);
+};
+
 module.exports = helpers;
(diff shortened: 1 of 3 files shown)


lib/helpers.js
```diff
@@ -149,4 +149,8 @@ helpers.loadConfigFile = function (configPath) {
   return file;
 };
 
+helpers.stripQuotes = function (str) {
```
DanPurdy: Are there tests for this coming on the develop branch?

DanPurdy:
Seems good, just the extra helper function tests missing..

bgriffith:
Tests added for helper function :rocket:

DanPurdy:
:+1:

PR #214, opened 2015-09-22
Hotfix Develop: cli output fix

DanPurdy:
same as #213 which should be merged first


DanPurdy:
#213 is merged so this should be good to go now

PR #217, opened 2015-09-22
Fix potential issue with no-transition-all

### LINKED ISSUE #218: no-transition-all rule can potentially cause error
The `no-transition-all` rule can potentially cause a fatal error if it discovers a declaration with no node of type ident.

bgriffith:
Add further checks to make sure that every declaration has both a property and an ident.

Fixes #218


DanPurdy:
Yeah glad we checked this :smile:

PR #232, opened 2015-09-28
Doc. Add link to property-sort-order config files

zallek:
https://github.com/sasstools/sass-lint/issues/234
These lists are useful for people being refactoring their base code. So a link to exhaustive lists of ordered properties in the documentation could be useful.

DCO 1.1 Signed-off-by: Nicolas Fortin <nicolas.fortin37@gmail.com>


DanPurdy:
Thanks for this @zallek would you mind having a quick read through the contribution guidelines and updating this PR with the required info.

I know it's only a small change but it's good to make sure we're all working to these

https://github.com/sasstools/sass-lint/blob/develop/CONTRIBUTING.md

Thanks again.

PR #233, opened 2015-09-28
Add rule: No mergeable selectors

### LINKED ISSUE #18: New Rule: Mergeable Selector
Check to see if selectors can be merged together

DanPurdy:
Add rule: no-mergeable-selectors which enforces the merging and nesting of selectors where possible.

There are two possible warning messages, the first is encountered if a selector is deemed mergeable:

> Rule `.image` should be merged with the rule on line 194

If a selector is deemed nestable the following error will occur:

> class .image should be nested within the parent class

There are multiple options available

`force-element-nesting` - enforces that all element selectors are nested where possible
`force-attribute-nesting` - enforces that all attribute selectors are nested where possible
`force-pseudo-nesting` - enforces that all pseudo elements/classes are nested where possible

`whitelist` - array of selector names you would like this rule to ignore and not enforce the merging of.

closes #18

On hold temporarily while I finish documentation and one last look over.


DanPurdy:
Seen as this was a massive rule in my opinion I've added a lot of comments. There's a lot going on which I'd like to take another look over once the docs are done. 
Added Jsdoc style comments for the functions created within too to make it easier for anyone picking this up in the future.

DanPurdy:
This is ready for review now.

Snugug:
So, I know you're going to hate me, because I'd hate me, but I feel as if this rule may be doing too much? If we stick to the small rules policy we've been pushing for, I feel as if `force-element-nesting`, `force-attribute-nesting`, and `force-pseudo-nesting` should all be their own rules, with mergeable selectors just focused on mergable selectors.

@bgriffith @DanPurdy what are your thoughts? If you both disagree with me we can keep as is (looks fine otherwise)

DanPurdy:
HA! So i actually knew this was probably going to be the thoughts (i had them too but the sweat and tears I poured into it made me sweep it under the rug!).

I do agree though that it's doing a LOT. If we do split these into separate rules I just felt like this would be duplicating functionality 3 or 4 times. Unless there's some way to run this master rule from 4 separate rules.. Do we think that's an option? Happy to split these up though as long as we're happy with the implementation

Snugug:
There are a couple of rules that we have that have very similar runtimes.

I think if we split the helpers you've made up in to `helpers.js` it may make it easier to deal with (and would give us the added benefit of being forced to write tests for them), but I also don't see a clean way to break this up.

One way we may be able to reduce the duplication is by having the sub rules' `define` simply return the main rule's `define` attribute with the new options set, but I don't know how those options would be retrieved.

DanPurdy:
Yeah that was what I was thinking, it's almost like a sub rule that just feeds options to the main rule. Definitely agree with the helper functions, will get those moved out. I'll create a PR for each rule too.

* mergeable selectors
* mergeable pseudo elements/classes
* mergeable attributes

Those 3 sound ok? Each will have a whitelist option too.

Maybe this is something we revisit when we look into external distributable rules / plugins? Just keep them split for now.

Snugug:
Yah, those all sound good to me.

DanPurdy:
Wait actually, I just re read your initial comment.. Forgive me..

This should be 4 rules then?

* `no-mergeable-selectors`
* `force-element-nesting`
* `force-attribute-nesting`
* `force-pseudo-nesting`

so basically just split the nesting out of this rule completely.

Snugug:
Yah, that. I thought you had 3 on top of mergeable selectors.

DanPurdy:
Sorry, confused myself too. I don't think the mergeable part needs to be split out.. they're either mergeable or not, no reason to split those between different selector groups.

DanPurdy:
Hmm some of these helper functions are relying on some of the AST's helper methods themselves.. Not seeing a good way of splitting those out really at the moment.

Snugug:
Is it possible to provide them with the correct AST context of split out?

DanPurdy:
When using them within the rule yes but we don't pass any AST context when we test them. The `constructSelector` funciton is the only one I think that would need it, everything else can be worked around.

Snugug:
Not at my computer, will look later tonight. Is that a _this_ bound function? If not, can we pass it to our stuff? If so, can we bind the _this_ to accommodate?

DanPurdy:
To a point I guess, It's using the `forEach` and `is` methods from `gonzales-pe`. Obviously `is` is easily replaced by just checking the type property but `forEach` is a different beast as it actually traverses multiple levels and is part of the `node.prototype` that `gonzales` provides. I don't think writing our own version is a good idea so that only leaves passing the full AST to our helpers test file. Had a quick look at that but not sure of a clean way really..

DanPurdy:
Going to close this issue and roll a fresh one to remove some of the outdated information surrounding this.

Issue #226, opened 2015-09-27
Allow Config Path to be assed as inline option


Snugug:
It occurs to me that the more options we're able to wrap up in our standard options, as opposed to needing to pass in additional arguments, the more future-friendly the library will be. Config Path is currently the only option that requires an additional argument. We should be able to define it in our options and check for it. I'd like to see this implemented in `options->config-file`

PR #244, opened 2015-10-04
Add Config File option for passed in options

Snugug:
Resolves sasstools/sass-lint#226


DanPurdy:
Looks fine to me. What is the use case for this though, I wasn't to clear on the original issue. You'd specify a global config file within a local config file?

Snugug:
For the task runners so they don't need to jump through hoops to have a user specify a custom config file. It's standardized this way.

PR #252, opened 2015-10-05
Fix Sass syntax issue with empty-line-between-block rule

### LINKED ISSUE #216: Sass Syntax - empty-line-between-blocks rule not working
The `empty-line-between-blocks` rule doesn't work with the Sass syntax. This is down to the parser interpreting the nodes differently.

SassLint: 1.2.2

bgriffith:
**NB: Work in progress.**

This PR fixes an issue with the `empty-line-between-block` rule and the Sass syntax. This has been a real pain mainly down to the differences in node structure that gonzales pumps out dependant on syntax. The end result isn't pretty and have resulted in using different code depending on which syntax is used.

Fixes #216 

### DIFF
--- a/lib/rules/empty-line-between-blocks.js
+++ b/lib/rules/empty-line-between-blocks.js
@@ -1,36 +1,92 @@
 'use strict';
 
-var helpers = require('../helpers'),
-    os = require('os');
+var helpers = require('../helpers');
+
+var counter,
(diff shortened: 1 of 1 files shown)


bgriffith:
Closing this PR as it is being replaced by #258 where the fix is being bundled for a 1.3 release.

Issue #17, opened 2015-08-27
New Rule: Name Format


Snugug:
Add rule to verify name format:

* Selectors
* Variables
* Functions
* Mixins

DanPurdy:
In the same vein as we discussed with merge able selectors, do we feel these are all separate rules @Snugug ? I think at the very least selectors should be separate to the other 3 but I'm wondering whether 4 rules is excessive before I start on this.

benthemonkey:
I'd like to have a shot at tackling this. For now, if everyone is ok with it, I'll have it parity the configuration options of scss-lint. Sound good, @Snugug @DanPurdy?

Snugug:
Go for it @benthemonkey 

We'd like to split the actual rule out, though, for each type, so `function-name-format`, `mixin-name-format`, `variable-name-format`, `placeholder-name-format`, etc…

DanPurdy:
Yeah go for it. I made a start but hadn't got too far, been a busy week! Definitely 4 separate rules though. Good luck!

benthemonkey:
Looking more carefully at both NameFormat and SelectorFormat from scss-lint, I completely agree with this decision. So the following rules will be added:
`function-name-format`
`mixin-name-format`
`variable-name-format`
`placeholder-name-format`

All having the same options:
`allow-leading-underscore`: boolean (default true)
`convention`: one of `hyphenatedlowercase` (default), `camelcase`, `snakecase`, or RegExp
`convention-explanation`: defaults to `should match regex /[regex]/`

DanPurdy:
Don't forget with parent selectors we should think about BEM style etc. I think the other formats should be ok but selector format will need another step or two!

benthemonkey:
I'm purposely biting off a manageable amount of code to add while I become familiar with contributing. So I've left off adding `selector-format` for now. Once these rules are added, I would like to attempt `selector-format`.

Thanks for the feedback though! I'll definitely keep it in mind.

DanPurdy:
Awesome, looking forward to it. If you need any help just let me know and I'll gladly sort something out!

DanPurdy:
@benthemonkey @bgriffith @Snugug Going to close this issue and setup 3 new separate issues for the remaining rules. Should help us to keep track of what is done and what is not.

benthemonkey:
@DanPurdy don't forget Placeholder Name Format

DanPurdy:
@benthemonkey Added that too now.

PR #256, opened 2015-10-06
Feature/mergeable selectors

DanPurdy:
Add rule: no-mergeable-selectors which enforces the merging of selectors where possible.

A warning/error message is returned as follows if a selector is deemed mergeable:

> Rule .image should be merged with the rule on line 194

There is a whitelist option available which prevents enforcement of this rule on specified selector patterns.
whitelist - array of selector names you would like this rule to ignore and not enforce the merging of.

closes #18

DCO 1.1 Signed-off-by: Dan Purdy danjpurdy@gmail.com

PR #257, opened 2015-10-07
Add rule: variable-name-format

benthemonkey:
Add rule: variable-name-format which validates that all variables adhere to a convention

Options:
`allow-leading-underscore`: Allow variables to begin with an underscore (default true)

I took the patterns for `isCamelCase`, `isHyphenatedLowercase` and `isSnakeCase` from how scss-lint implements them for consistency between projects.

I'll add the related rules to this Pull Request, but since they are very similar I wanted to get feedback early to avoid applying fixes across four rules.

Part 1 of 4 for #17 

### DIFF
--- a/docs/rules/variable-name-format.md
+++ b/docs/rules/variable-name-format.md
@@ -5,12 +5,16 @@ Rule `variable-name-format` will enforce the use of hexadecimal color values rat
 ## Options
 
 * `allow-leading-underscore`: `true`/`false` (defaults to `true`)
-* `convention`: `'hyphenatedlowercase'` (default), `camelcase`, `snakecase`, or a Regular Expression that the variable name must match (e.g. `[_A-Z]`)
+* `convention`: `'hyphenatedlowercase'` (default), `camelcase`, `snakecase`, or a Regular Expression that the variable name must match (e.g. `/^[_A-Z]+$/`)
 * `convention-explanation`: Custom explanation to display to the user if a variable doesn't adhere to the convention
 
-## Examples
+## Example 1
 
-When enabled (assuming `allow-leading-underscore: true` and `convention: hyphenatedlowercase`), the following are allowed:
+Settings:
+- `allow-leading-underscore: true`
+- `convention: hyphenatedlowercase`
+
+When enabled, the following are allowed:
 
 ```scss
 $hyphenated-lowercase: 1px;
@@ -22,7 +26,7 @@ $_leading-underscore: 1px;
 
 ```
 
-When enabled (assuming `allow-leading-underscore: true` and `convention: hyphenatedlowercase`), the following are disallowed:
+When enabled, the following are disallowed:
 
 ```scss
 $HYPHENATED-UPPERCASE: 1px;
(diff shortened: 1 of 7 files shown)


DanPurdy:
Looks good, only thing I feel missing is a hyphenated string in the `hyphenated lowercase` helper tests. Also the documentation file but otherwise I'm happy with it! 

Thanks @benthemonkey

DanPurdy:
Also let's keep the pull requests separate for each rule. Makes it a bit easier for us to keep track of the changes when we do a release. That ok?

bgriffith:
Hey @benthemonkey, thanks for the PR. Could you make sure to link to the issues so it's easier to see what this is related to? Thanks!

DanPurdy:
Awesome, this looks great.

So i know the change to `Sass` syntax wont affect this much but would you mind adding test for `.sass` too.

#256 and #258 both have examples of us providing these, we just want to cover all bases moving forward! If you wouldn't mind just adding these and then I'll happily merge this in! 

Thanks again

DanPurdy:
Also, in the documentation maybe a quick demonstration and example of each of the other possible options as they are specified options. Just to be thorough. @Snugug @bgriffith what do you guys reckon?

benthemonkey:
Why do we only count the number of warnings in test cases? Isn't that prone to missing issues?

Wouldn't the library be more robust if each unit test looked at the smallest amount of sass/scss to produce a warning, and have the sass/scss directly in the test instead of in a file?

Like this (could be significantly cleaned up with some helper functions):

```javascript
describe('variable name format', function () {
  describe('[convention: hyphenatedlowercase]', function () {
    lint.test(
      {
        'text': '$hyphenated-lower-case: 1;',
        'format': 'scss',
        'filename': 'none'
      },
      {
        'variable-name-format': [
          1,
          {
            'convention': 'camelcase'
          }
        ]
      },
      function (data) {
        it('correct warning count', function () {
          lint.assert.equal(1, data.warningCount);
        });

        it('contains variable name in message', function () {
          lint.assert.ok(data.messages[0].message.indexOf('hyphenated-lower-case'));
        });

        it('contains correct convention in message', function () {
          lint.assert.ok(data.messages[0].message.indexOf('camelCase'));
        });
      }
    );
  });
});
```

DanPurdy:
This is definitely something that should be looked at and added and something that @bgriffith and I for certain have spoken about and I'm sure @Snugug is the same. I feel it works alongside what we have not instead of too.

My simple answer would be, it was like this when I got here. :smile:  My slightly more long winded answer is that we've found time and time again with some of the rules in the early days that the context of what was being tested i.e. nested rules etc affected certain other contexts or caused outright failures, so in essence we are making sure that the amount of errors we have designed to be reported are. There was a bug in `mixins-before-declarations` for example that only reared it's head when it followed another rule block, which without being EXTREMELY thorough would probably have been missed by small single line/block Sass/scss tests. The other issue is some of the parsing errors that our AST occasionally throws up again being sometimes fairly context specific.

I definitely think your suggestion of actually counting the instances of an error message drastically increases the reliability of the tests and would be a welcome addition and I also feel like i'd love to see some full scale tests too with a 'real-world' file being linted with all the rules active and everything being checked for interoperability and any clashes or conflicts.

Let's create an issue from this though and look to improve them as we move forward, the more tests the better.

DanPurdy:
:+1:

Issue #166, opened 2015-09-13
Sass Syntax - Tests progress


bgriffith:
Having had a few Sass Syntax related issues popping up, I think it would be a good move to create tests for that syntax too.

I've started going through the current tests and have created quite a few .sass equivalents. So far there has been a few tests that have failed but the majority are ok.

By testing both the SCSS and Sass syntax I feel we will then be able to confidently say that we have cross syntax support.

I'll update the list below as I go.

### The rules

- [x] border-zero
- [x] clean-import-paths - (Issue #179, Fixed #205)
- [x] empty-args
- [x] empty-line-between-blocks - (:bug: Issue #216)
- [x] extends-before-declarations
- [x] extends-before-mixins - (Fixed issue #203)
- [x] final-newline - (Issue #207, Fixed #208)
- [x] hex-length
- [x] hex-notation
- [x] indentation
- [x] leading-zero
- [x] mixing-before-declarations
- [x] nesting-depth
- [x] no-color-keywords
- [x] no-color-literals (Issue #186, Fixed in #187)
- [x] no-css-comments
- [x] no-debug
- [x] no-duplicate-properties
- [x] no-extends - (Fixed issue - #173)
- [x] no-ids
- [x] no-important
- [x] no-invalid-hex
- [x] no-qualifying-elements
- [x] no-trailing-zeros
- [x] no-url-protocols
- [x] no-vendor-prefix (Issue #182, Fixed in #183)
- [x] no-warn
- [x] placeholder-in-extend - (Fixed in #204)
- [x] property-sort-order
- [x] quotes
- [x] shorthand values
- [x] single-line-per-selector - (Fixed issue - #165)
- [x] space-after-bang
- [x] space-after-colon
- [x] space-after-comma
- [x] space-before bang
- [x] space-before-colon
- [x] space-before-parens
- [x] url-quotes
- [x] variable-for-property
- [x] zero-unit

### Rules that are N/A
- trailing-semicolon
- space-before-brace
- no-empty-rulesets
- brace-style
- one-declaration-per-line

DanPurdy:
Awesome, I started having a look at this after we spoke this morning, but looks like you've got further ahead than me. Do you wanna explain a little how you're going about it, I've got a few rules nearly done here that i'd like to get the sass tests up and running for too, be good to have them in the same format as yours.

bgriffith:
Sure, it's very simple and happy to change if you've got a better suggestion.

[...]

PR #258, opened 2015-10-07
Add Sass tests for rules & fix related issues

### LINKED ISSUE #216: Sass Syntax - empty-line-between-blocks rule not working

### LINKED ISSUE #228: empty-line-between-blocks should not raise warning when comments exist
Currently comments on the line prior to the first nested selector in a block causes an empty-line-between-blocks warning to be raised. This comment should be treated as a non-existant line and should not raise a warning.

To reproduce, add the following to the bottom of [empty-line-between-blocks.scss](https://github.com/sasstools/sass-lint/blob/develop/tests/sass/empty-line-between-blocks.scss):

```scss
.block {
  // Comment to describe complexity
  .element {
    content: 'baz';
  }
}

.block {
  /* Comment to describe complexity */
  .element {
    content: 'baz';
  }
}
```

I would expect that this would raise no additional linting errors.

---

Interestingly this currently does not occur if the first item within the block (after the comment) is a mixin, extends or media query. The following all pass as expected:

```scss
.block {
  // Comment to describe complexity
  @include foo;
}

.block {
  // Comment to describe complexity
  @extend .foo;
}

.block {
  // Comment to describe complexity
  @media (min-width:50em) {
    content: 'qux';
  }
}
```

bgriffith:
Apologies for the size of this PR.

It adds Sass tests for the majority of the rules listed here #166. The only one missing is indentation as this needs a rewrite.

It also includes a fix for the `empty-line-between-block` rule as this required helpers bundled with the soon to be released version 1.3 on develop which will close #216 and close #228 


tests/sass/no-invalid-hex.sass
```diff
@@ -0,0 +1,45 @@
...
+
+$invalid-mixed-long: #1bcdefg
+$invalid-mixed-med: #1bcd
+$invalid-mixed-short: #1b
+
+$invalid-mixed-letters-long: #abcdef7
+$invalid-mixed-letters-med: #abc4
+$invalid-mixed-letters-short: #a1
+
+$invalid-character-map: (invalid-characters-upper-letters: #GHIJKL, invalid-characters-upper-letters-short: #GHI, even-more-invalid-map: ( invalid-characters-lower-letters-short: #ghijkl, invalid-characters-lower-letters-short: #ghi ))
```
DanPurdy: Are you forced to write a map on a single line in Sass?

bgriffith: Yeah... it's horrendous.

tests/sass/mixins-before-declarations.sass
```diff
@@ -0,0 +1,46 @@
...
+
+
+    +waldo
+
+    &--modifier
+      +foo('yo')
+      +hello
+      +test
+      height: 100px
+      +test-again
```
DanPurdy: There are some tests missing here from the develop branch version. I added this as part of a bug fix, would be worth just double checking any extra tests we have added are added to these too.

tests/sass/extends-before-declarations.sass
```diff
@@ -0,0 +1,29 @@
...
+
+  .foo
+    content: 'where'
+
+    @extend %waldo
+
+
+  .bar
+    @extend %waldo
+    content: 'where'
```
DanPurdy: Newline at end of file missing here?

tests/rules/space-before-brace.js
```diff
@@ -2,9 +2,12 @@
 
 var lint = require('./_lint');
 
-var file = lint.file('space-before-brace.scss');
+//////////////////////////////
```
DanPurdy: I feel this rule should probably provide a warning itself if enabled on a Sass syntax file? What do you reckon?

tests/sass/extends-before-declarations.sass
bgriffith: Added :+1:

tests/rules/space-before-brace.js
Snugug: Other rules that don't apply to Sass skip silently and I think we should stick to that in case someone has a mix of sass and sass

tests/sass/mixins-before-declarations.sass
bgriffith: Added too, though it causes a parse error so commented some out for now with `FIXME` comment

tests/rules/space-before-brace.js
DanPurdy: Awesome, never saw that.

tests/rules/indentation.js
```diff
@@ -17,3 +17,21 @@ describe('indentation', function () {
     });
   });
 });
+
+
+//////////////////////////////
```
DanPurdy: Lets not forget about these in `1.4.0` :smile:

lib/rules/empty-line-between-blocks.js
```diff
@@ -2,34 +2,96 @@
...
+    syntax;
+
 var findNearestReturn = function (parent, i) {
   var previous,
       doublePrevious,
       space;
 
-  if (i >= 2) {
-    doublePrevious = parent.content[i - 2];
+  if (syntax === 'scss') {
```
DanPurdy: Seen as you're already performing this checking in the rule itself would it not be better to split these into 2 separate functions? I feel this one is doing too much on it's own.

lib/rules/empty-line-between-blocks.js
```diff
@@ -41,30 +103,90 @@ module.exports = {
...
-    ast.traverseByType(['ruleset'], function (ruleset, i, parent) {
+    // Set syntax
+    syntax = ast.syntax;
+
+    ast.traverseByType('ruleset', function (node, j, p) {
       var space;
 
-      space = findNearestReturn(parent, i);
+      if (syntax === 'scss') {
+        space = findNearestReturn(p, j);
```
DanPurdy: When the returned object contains a property named space I think this should probably be called something else.. `space.space` is a little confusing.

DanPurdy:
Great stuff. This should make sass-lint even more bullet proof moving forward, although this hides the fact that you've actually fixed a lot of these rules in hotfixes already so thanks for those!

My only other comment is maybe to add a few more comments as in the `no-mergeable-selector` rules earlier, it's fairly tough to trace the reasoning for certain aspects of this rule.

bgriffith:
This `empty-line-between-rule` has still got some issues on the SCSS side so will fix before we merge this. See comment: https://github.com/sasstools/sass-lint/issues/228#issuecomment-146500067

DanPurdy:
Awesome!

PR #265, opened 2015-10-08
Fix a bug in shorthand-values

### LINKED ISSUE #263: Issue with short-hand values and values of 0
The `shorthand-values` rule doesn't work with values of zero.

Best way to replicate this is to add the following to the tests:

```scss
.value-of-zero {
  border-radius: 0;
  border-width: 0;
  margin: 0;
  padding: 0;
}
```

It currently returns false warnings for all of these.

screen shot 2015-10-08 at 16 51 36

Sass-lint: dev@latest

DanPurdy:
Fixes a bug when using plain number values such as `0` in the `shorthand-values` rule.

Only merging to develop due to the impending 1.3.0 release

fixes #263 

### DIFF
--- a/lib/rules/shorthand-values.js
+++ b/lib/rules/shorthand-values.js
@@ -82,15 +82,18 @@ module.exports = {
                   fullVal += el.content;
                 });
               }
-              else if (val.is('operator') || val.is('ident')) {
+
+              else if (val.is('operator') || val.is('ident') || val.is('number')) {
                 fullVal += val.content;
               }
+
               else if (val.is('variable')) {
                 val.forEach(function (el) {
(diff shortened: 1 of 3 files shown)


bgriffith:
Does it work with just `margin: 0`? Probably good idea to just add some more tests.

bgriffith:
Ah suppose you have it with `border-radius: 0`

DanPurdy:
It works exactly the same for any of the following properties

```
'border-color',
  'border-radius',
  'border-style',
  'border-width',
  'margin',
  'padding'
```

bgriffith:
Sounds good

PR #269, opened 2015-10-08
Rule: Force element nesting

### LINKED ISSUE #266: New Rule: Force element nesting
As discussed in #256 

Rule should enforce the nesting of elements

DanPurdy:
Adds the rule `force-element-nesting` which does exactly what it says on the tin and forces the nesting of elements.

The warning message:

> class `.test` should be nested within its parent selector

Decided upon in #256 

closes #266

I would really appreciate it if anyone can think of any extra scenarios to test here.


tests/rules/force-element-nesting.js
```diff
@@ -0,0 +1,35 @@
...
+  });
+});
+
+//////////////////////////////
+// Sass syntax tests
+//////////////////////////////
+describe('force element nesting - sass', function () {
+  var file = lint.file('force-element-nesting.sass');
+
+  it('[convention: 0]', function (done) {
```
bgriffith: There's no option for convention

PR #270, opened 2015-10-08
Rule: Force attribute nesting

### LINKED ISSUE #267: New Rule: Force attribute nesting
As discussed in #256

Rule should enforce the nesting of attributes

DanPurdy:
Adds the rule `force-attribute-nesting` which does exactly what it says on the tin and forces the nesting of elements.

The warning message:

> attribute .[type='text'] should be nested within its parent selector

Decided upon in #256

closes #267

As with #269 I would really appreciate it if anyone can think of any extra scenarios to test here.


docs/rules/force-attribute-nesting.md
```diff
@@ -0,0 +1,47 @@
...
+}
+
+a {
+  &[target='_blank'] {
+    content: '';
+  }
+}
+
+.form {
+  .class input{
```
bgriffith: Space before `{`

tests/rules/force-attribute-nesting.js
```diff
@@ -0,0 +1,35 @@
...
+  });
+});
+
+//////////////////////////////
+// Sass syntax tests
+//////////////////////////////
+describe('force attribute nesting - sass', function () {
+  var file = lint.file('force-attribute-nesting.sass');
+
+  it('[convention: 0]', function (done) {
```
bgriffith: No convention option?

PR #274, opened 2015-10-09
Fix final newline rule and Sass syntax

### LINKED ISSUE #273: Sass syntax - final newline rule issues
Managed to replicate an issue where the final newline rule doesn't work with sass syntax as noted here https://github.com/sasstools/sass-lint/issues/216#issuecomment-144994176 by @za-creature.

bgriffith:
There was an issue with Sass syntax and final newline rule not working. This PR fixes #273.


tests/sass/final-newline--return.sass
```diff
@@ -1,4 +1,2 @@
 .foo
   content: bar
-
```
DanPurdy: Did you mean to remove both of these?

DanPurdy:
:+1:

PR #281, opened 2015-10-09
Add rule: mixin-name-format

### LINKED ISSUE #276: New Rule: Mixin Format
Add rule to verify name format of mixins.

Please see the original issue for more information and discussion #17

benthemonkey:
Add rule: variable-name-format which validates that all variables adhere to a convention

Options:
`allow-leading-underscore`: Allow variables to begin with an underscore (default true)

Closes #276

### DIFF
--- a/lib/rules/mixin-name-format.js
+++ b/lib/rules/mixin-name-format.js
@@ -1,0 +1,78 @@
+// Note that this file is nearly identical to function-name-format.js, variable-name-format.js, and placeholder-name-format.js
+'use strict';
+
+var helpers = require('../helpers');
+
+module.exports = {
+  'name': 'mixin-name-format',
+  'defaults': {
+    'allow-leading-underscore': true,
+    'convention': 'hyphenatedlowercase'
+  },
+  'detect': function (ast, parser) {
+    var result = [];
+
+    ast.traverseByTypes(['mixin', 'include'], function (node) {
+      var name;
+
+      if (node.type === 'mixin') {
+        name = node.first('ident').content;
+      }
+      else {
+        name = node.first('simpleSelector').first('ident').content;
+      }
+
+      var originalName = name;
+
+      if (parser.options['allow-leading-underscore'] && name[0] === '_') {
+        name = name.slice(1);
+      }
+
+      if (parser.options.convention === 'hyphenatedlowercase' && !helpers.isHyphenatedLowercase(name)) {
+        result = helpers.addUnique(result, {
+          'ruleId': parser.rule.name,
+          'line': node.start.line,
+          'column': node.start.column,
+          'message': 'Mixin \'' + originalName + '\' should be written in lowercase with hyphens',
+          'severity': parser.severity
+        });
+      }
+      else if (parser.options.convention === 'camelcase' && !helpers.isCamelCase(name)) {
+        result = helpers.addUnique(result, {
+          'ruleId': parser.rule.name,
+          'line': node.start.line,
+          'column': node.start.column,
+          'message': 'Mixin \'' + originalName + '\' should be written in camelCase',
+          'severity': parser.severity
+        });
+      }
+      else if (parser.options.convention === 'snakecase' && !helpers.isSnakeCase(name)) {
+        result = helpers.addUnique(result, {
+          'ruleId': parser.rule.name,
+          'line': node.start.line,
+          'column': node.start.column,
+          'message': 'Mixin \'' + originalName + '\' should be written in snake_case',
+          'severity': parser.severity
+        });
+      }
+      else if (parser.options.convention instanceof RegExp && !parser.options.convention.test(name)) {
+        var message = parser.options['convention-explanation'];
+
+        if (!message) {
+          message = 'should match regular expression /' + parser.options.convention.source + '/' +
+                    (parser.options.convention.ignoreCase ? 'i' : '');
+        }
+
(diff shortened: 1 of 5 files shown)


lib/rules/mixin-name-format.js
```diff
@@ -0,0 +1,78 @@
...
+      var name;
+
+      if (node.type === 'mixin') {
+        name = node.first('ident').content;
+      }
+      else {
+        name = node.first('simpleSelector').first('ident').content;
+      }
+
+      var originalName = name;
```
DanPurdy: Can we declare all variables at the top of the block please.

DanPurdy: I would also change the order here maybe..

I would probably change the order here a little for mainly flow reasons.

```
var originalName,
      name;

      if (node.type === 'mixin') {
        originalName = node.first('ident').content;
      }
      else {
        originalName = node.first('simpleSelector').first('ident').content;
      }

      name = originalName;

      if (parser.options['allow-leading-underscore'] && name[0] === '_') {
        name = name.slice(1);
      }
```

Almost identical but feels a little more consistent in flow. Ignore me if you want though :smile:

benthemonkey: Updated. I didn't like the variable `originalName` very much so I went with `name` and `strippedName`.

benthemonkey:
Once this is merged, I'll start using feature branches instead of using the develop branch on my fork. Is there anything else I should change?

lib/rules/mixin-name-format.js
```diff
@@ -0,0 +1,79 @@
...
+        result = helpers.addUnique(result, {
+          'ruleId': parser.rule.name,
+          'line': node.start.line,
+          'column': node.start.column,
+          'message': 'Mixin \'' + name + '\' should be written in snake_case',
+          'severity': parser.severity
+        });
+      }
+      else if (parser.options.convention instanceof RegExp && !parser.options.convention.test(strippedName)) {
+        var message = parser.options['convention-explanation'];
```
DanPurdy: `convention-explanation` isn't declared by default, if someone were to include their own regex convention but not include an explanation this would be undefined? Not really great to check against undefined so should probably declare it in the defaults above.

lib/rules/mixin-name-format.js
```diff
@@ -0,0 +1,79 @@
...
+        result = helpers.addUnique(result, {
+          'ruleId': parser.rule.name,
+          'line': node.start.line,
+          'column': node.start.column,
+          'message': 'Mixin \'' + name + '\' should be written in lowercase with hyphens',
+          'severity': parser.severity
+        });
+      }
+      else if (parser.options.convention === 'camelcase' && !helpers.isCamelCase(strippedName)) {
+        result = helpers.addUnique(result, {
```
DanPurdy: Just a thought but this setting of result doesn't need to be repeated for each convention or case. You could set an error flag and then set the message here and then check for the error flag after all the checks at the end meaning you'd only need to write create a result once..

```js
if (parser.options.convention === 'hyphenatedlowercase' && !helpers.isHyphenatedLowercase(strippedName)) {
  errorFlag = true;
  message = 'lowercase with hyphens';
}

......

if (errorFlag) {
          'ruleId': parser.rule.name,
          'line': node.start.line,
          'column': node.start.column,
          'message': 'Mixin \'' + name + '\' should be written in ' + message
          'severity': parser.severity
        });
}
```

DanPurdy:
I added a few bits that might help streamline this a bit but they're only minor in my view. As for if there's anything else you need to change.. Definitely not, it's great to have more people contributing and we're really grateful that you are! Thanks!

benthemonkey:
Thanks for the code suggestions. Since the structure is nearly identical to variable-name-format, do I commit the change to that file in this PR?

DanPurdy:
Let's keep it to separate PR if you don't mind, you can use this PR's id as a reference though as an issue would be overkill in this instance.

benthemonkey:
Do you have any opinion about whether or not I should change the functionality of convention-explanation so that it overrides the message for any convention, and not just the regular expression option (this is how scss-lint does it)?

DanPurdy:
Personally I think keep it as you have it. 

We could accidentally provide a way for our users to become very confused if a convention message is left in their config to say don't use this regex when it's actually warning against snake case for example. But then I've always thought that the extra explanation option in scss-lint was a strange one and also a one off and would of removed it all together :smile: Seems strange for only this rule to allow it IMO.

benthemonkey:
I realized that js-yaml doesn't read in RegExp by default, so (while the tests didn't cover this) I need to update the rule to properly handle Regular Expressions (and this will certainly require an update to the variable-name-format rule in another PR)

DanPurdy:
So are you happy for this to go through now? It looks good to me. If you could fix the variable-name-format rule next and then we'll move function, placeholder and selector formats into `1.4.0`?

benthemonkey:
Yeah I'm happy with this rule now.

I don't know what the plan is for releasing 1.3, but I could submit PRs for variable-name-format, function and placeholder by end of day today. Selector certainly will be more work.

DanPurdy:
Ok cool, we're planning on getting `1.3.0` out tomorrow now I believe. so that would be great if you could.

Thanks again for this.

benthemonkey:
Darn! I just saw one place I forgot to fix! The unit test for sass with regexp...

Issue #260, opened 2015-10-08
Multi-line function/mixin calls should be acceptable when enforcing space-between-parens and indentation


BPScott:
Sometimes the argument list when calling functions/mixins ends up being quite long so it is nice to put them onto multiple lines to make it them more readable. However currently this throws a combination of warnings - two `space-between-parens` (one for the opening paren, one for the closing paren) and an `indentation` warning for every line within the arguments. This occurs for native CSS functions (e.g. linear-gradient()), user-defined functions and mixins. To see this in action, run the following scss with the ruleset `{'indentation': 1, 'space-between-parens': 1}`:

```scss
// Top-level mixin
@include hello(
  $foo,
  $bar,
  $baz
);

.foo {
  // Nested Mixin
  @include hello(
    $foo,
    $bar,
    $baz
  );
  
  // CSS built-in function
  background: transparent linear-gradient(
    to bottom,
    #ff0000 0%,
    #00ff00 40%,
    #0000ff 40%
  );

  // User-defined function
  content: goodbye(
    $foo,
    $bar
  );
}
```

I would expect this to raise no warnings.

bgriffith:
Have this fixed locally. Will create PR first thing tomorrow.

PR #284, opened 2015-10-09
Allow multi-line arguments with space-between-parens rule

bgriffith:
This PR fixes the false positive that would arise when it came across multi-line arguments #260.


DanPurdy:
:+1: 

Fixes #260

PR #285, opened 2015-10-09
Ignore single line rulesets with empty-line-between-blocks rule

### LINKED ISSUE #282: empty-line-between-blocks should have an option to ignore single line blocks
Sometimes when writing many short blocks where there is a single property (/ mixin / extend) per block it is useful to group them close to each other as they are often inherently related. For instance a set of width classes can be densely packed instead of having a blank line between each class:

```scss
.w1-of-12 { width: (1/12) * 100%; }
.w2-of-12 { width: (2/12) * 100%; }
.w3-of-12 { width: (3/12) * 100%; }
.w4-of-12 { width: (4/12) * 100%; }
.w5-of-12 { width: (5/12) * 100%; }
.w6-of-12 { width: (6/12) * 100%; }
.w7-of-12 { width: (7/12) * 100%; }
.w8-of-12 { width: (8/12) * 100%; }
.w9-of-12 { width: (9/12) * 100%; }
.w10-of-12 { width: (10/12) * 100%; }
.w11-of-12 { width: (11/12) * 100%; }
.w12-of-12 { width: (12/12) * 100%; }
```

[Scss-lint's empty-line-between-blocks rule](https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#emptylinebetweenblocks) allows these single-line rule sets to sit next to each other as it has a configuration option `allow_single_line_rule_sets` that defaults to `true`. The configuration option may be set to off to disallow this exception.

Sass-lint should mimic this behaviour so that I may write the above without raising warnings.

bgriffith:
As discussed in #282 - this PR adds the option to ignore single line rulesets from the rule with the option `ignore-single-line-rulesets` and the default value being set to `true`.

This fixes #282 

### DIFF
--- a/docs/rules/empty-line-between-blocks.md
+++ b/docs/rules/empty-line-between-blocks.md
@@ -5,21 +5,24 @@ Rule `empty-line-between-blocks` will enforce whether or not nested blocks shoul
 ## Options
 
 * `include`: `true`/`false` (defaults to `true`)
+* `ignore-single-line-rulesets`: `true`/`false` (defaults to `true`)
 
 ## Examples
+
+### `include`
 
 When `include: true`, the following are allowed. When `include: false`, the following are disallowed:
 
 ```scss
 .foo {
-  content: 'bar';
+  content: 'foo';
 
-  .baz {
-    content: 'qux';
+  .bar {
+    content: 'bar';
 
     // Waldo
-    &-- {
-      content: 'alpha';
+    &--baz {
+      content: 'baz';
     }
   }
 }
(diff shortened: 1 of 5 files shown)


docs/rules/empty-line-between-blocks.md
```diff
@@ -29,13 +32,34 @@ When `include: false`, the following are allowed. When `include: true`, the foll
...
+
+When `ignore-single-line-rulesets: true`, the following are allowed. When `ignore-single-line-rulesets: false`, the following are disallowed:
+
+```scss
+.foo { content: 'foo'; }
+.bar { content: 'bar'; }
+.baz { content: 'baz'; }
+```
+
+When `ignore-single-line-rulesets: false`, the following are allowed. When `ignore-single-line-rulesets: true`, the following are disallowed:
```
DanPurdy: These have lines between them so if true would it actually complain? when set to true? I dont think it should.. You should still be allowed to have lines between single line blocks even if you choose to ignore them here.

DanPurdy:
Other than the possible clarification in the docs this good to go.

lib/rules/empty-line-between-blocks.js
```diff
@@ -105,7 +105,8 @@ var findNearestReturnSass = function (parent, i) {
 module.exports = {
   'name': 'empty-line-between-blocks',
   'defaults': {
-    'include': true
+    'include': true,
+    'ignore-single-line-rulesets': true
```
DanPurdy: can we change this to `allow-single-line-rulesets` as per the latest comments in #282 ?

bgriffith:
Updated

PR #293, opened 2015-10-11
Hotfix/no duplicate properties

### LINKED ISSUE #280: no-duplicate-properties should allow repeat properties when they are directly next to each other
Duplicating properties is sometimes required - such as when providing fallbacks for unsupported features, e.g. the following is a legitimate usage but currently raises a warning:

```scss
.foo {
  display: inline-block;
  display: flex;
}
```

Sass-lint should follow [CSSLint's lead](https://github.com/CSSLint/csslint/wiki/Disallow-duplicate-properties) and allow duplicate properties when they have different values but are immediately next to each other. I.e. the following are warnings:

```scss
/* properties with the same value */
.mybox {
  border: 1px solid black;
  border: 1px solid black;
}

/* properties separated by another property */
.mybox {
  border: 1px solid black;
  color: green;
  border: 1px solid red;
}
```

But the following should be acceptable, and not raise a warning:

```scss
/* one after another with different values */
.mybox {
  border: 1px solid black;
  border: 1px solid red;
}
```

DanPurdy:
This PR updates the no-duplicate-properties rule as discussed in #280 

Excluded duplicate properties must now immediately follow each other when used.

Extra tests are added to check for these new cases.

Documentation is also updated to add examples.

Closes #280

### DIFF
--- a/tests/rules/no-duplicate-properties.js
+++ b/tests/rules/no-duplicate-properties.js
@@ -33,7 +33,7 @@ describe('no duplicate properties - scss', function () {
     });
   });
 
-  it('enforce - [exclude: background]', function (done) {
+  it('enforce - [exclude: background, display]', function (done) {
     lint.test(file, {
       'no-duplicate-properties': [
         1,
@@ -82,7 +82,7 @@ describe('no duplicate properties - sass', function () {
     });
   });
 
-  it('enforce - [exclude: background]', function (done) {
+  it('enforce - [exclude: background, display]', function (done) {
     lint.test(file, {
       'no-duplicate-properties': [
         1,
--- a/tests/rules/no-duplicate-properties.js
+++ b/tests/rules/no-duplicate-properties.js
@@ -12,7 +12,7 @@ describe('no duplicate properties - scss', function () {
     lint.test(file, {
       'no-duplicate-properties': 1
     }, function (data) {
-      lint.assert.equal(4, data.warningCount);
+      lint.assert.equal(6, data.warningCount);
       done();
     });
   });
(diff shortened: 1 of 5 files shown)


tests/rules/no-duplicate-properties.js
```diff
@@ -28,7 +28,24 @@ describe('no duplicate properties - scss', function () {
         }
       ]
     }, function (data) {
-      lint.assert.equal(3, data.warningCount);
+      lint.assert.equal(5, data.warningCount);
+      done();
+    });
+  });
+
+  it('enforce - [exclude: background]', function (done) {
```
bgriffith: Should be `[exclude: background, display]`

tests/rules/no-duplicate-properties.js
```diff
@@ -60,7 +77,24 @@ describe('no duplicate properties - sass', function () {
...
```
bgriffith: Should be `[exclude: background, display]`

tests/sass/no-duplicate-properties.scss
```diff
@@ -20,3 +24,16 @@
...
+  background: rgb(200, 54, 54);
+  content: 'dec between background';
+  background: rgba(200, 54, 54, 0.5);
+}
+
+.display {
+  content: 'display';
+  display: flex;
+  display: inline-block;
+
```
bgriffith: Is that whitespace intentional?

DanPurdy: Yeah, to make sure it doesn't error on empty lines.

PR #294, opened 2015-10-11
Add rule: function-name-format

### LINKED ISSUE #277: New Rule: Function Format
Add rule to verify name format of functions.

Please see the original issue for more information and discussion #17

benthemonkey:
Add rule: function-name-format which validates that all functions adhere to a convention

Options:
`allow-leading-underscore`: Allow functions to begin with an underscore (default true)

Closes #277 

I'd like to note that I feel dirty copy/pasting this much code...

### DIFF
--- a/docs/rules/function-name-format.md
+++ b/docs/rules/function-name-format.md
@@ -1,0 +1,122 @@
+# Function Name Format
+
+Rule `function-name-format` will enforce a convention for function names.
+
+## Options
+
+* `allow-leading-underscore`: `true`/`false` (defaults to `true`)
+* `convention`: `'hyphenatedlowercase'` (default), `camelcase`, `snakecase`, or a Regular Expression that the function name must match (e.g. `^[_A-Z]+$`)
+* `convention-explanation`: Custom explanation to display to the user if a function doesn't adhere to the convention
+
+## Example 1
+
+Settings:
+- `allow-leading-underscore: true`
+- `convention: hyphenatedlowercase`
+
+When enabled, the following are allowed:
+
+```scss
+$foo: hyphenated-lowercase();
+$foo: _leading-underscore();
+
+.foo {
+  content: hyphenated-lowercase();
+}
+
+```
+
+When enabled, the following are disallowed:
+
+```scss
+$foo: HYPHENATED-UPPERCASE();
+$foo: _camelCaseWithLeadingUnderscore();
+
+.foo {
+  content: snake_case();
+}
+```
+
+## Example 2
+
+Settings:
+- `allow-leading-underscore: false`
+- `convention: camelcase`
+
+When enabled, the following are allowed:
+
+```scss
+$foo: camelCase();
+
+.foo {
+  content: anotherCamelCase();
+}
+```
+
+When enabled, the following are disallowed:
+
+```scss
+$foo: HYPHENATED-UPPERCASE();
+$foo: _camelCaseWithLeadingUnderscore();
+
+.foo {
+  content: snake_case();
+}
+```
+
+## Example 3
+
+Settings:
+- `allow-leading-underscore: false`
+- `convention: snakecase`
+
+When enabled, the following are allowed:
+
+```scss
+$foo: snake_case();
+
+.foo {
+  content: another_snake_case();
+}
+```
+
+When enabled, the following are disallowed:
+
+```scss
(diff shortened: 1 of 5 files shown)


tests/sass/function-name-format.scss
```diff
@@ -0,0 +1,12 @@
+$foo: kabab-case();
+$foo: snake_case();
+$foo: camelCase();
+$foo: PascalCase();
+$foo: Camel_Snake_Case();
+$foo: SCREAMING_SNAKE_CASE();
+$foo: _with-leading-underscore();
+$foo: _does_NOT-fitSTANDARD();
+
```
DanPurdy: Would be worth adding a function declaration into these just to make sure we're catching those in the tests going forward.

```scss
@function cp($target, $container) {
  @return calc-percent($target, $container);
}
```

benthemonkey: I must stink at reading Sass' documentation because I never realized this was a thing!

benthemonkey: can you find the Sass equivalent for your example? I can't figure out the right syntax.

DanPurdy: You can just do the following

```sass
@function cp($target, $container)
  @return calc-percent($target, $container)
```

docs/rules/function-name-format.md
```diff
@@ -0,0 +1,122 @@
...
+$foo: SCREAMING_SNAKE_CASE();
+
+.foo {
+  content: _LEADING_UNDERSCORE();
+}
+```
+
+When enabled, the following are disallowed:
+
+(Each line with a variable will report `Variables must contain only uppercase letters and underscores` when linted.)
```
DanPurdy: this should be 'functions must... '

docs/rules/function-name-format.md
```diff
@@ -0,0 +1,122 @@
...
+  content: camelCase();
+}
+```
+
+## Example 4
+
+Settings:
+- `allow-leading-underscore: true`
+- `convention: '^[_A-Z]+$'`
+- `convention-explanation: 'Variables must contain only uppercase letters and underscores'`
```
DanPurdy: Should be 'functions must... '

lib/rules/function-name-format.js
```diff
@@ -0,0 +1,67 @@
...
+          violationMessage = false;
+
+      strippedName = name;
+
+      if (parser.options['allow-leading-underscore'] && name[0] === '_') {
+        strippedName = name.slice(1);
+      }
+
+      switch (parser.options.convention) {
+      case 'hyphenatedlowercase':
```
DanPurdy: Although `npm test` and eslint aren't flagging this as an error technically these cases and default should be indented within the switch.

docs/rules/function-name-format.md
```diff
@@ -0,0 +1,122 @@
...
+## Example 1
+
+Settings:
+- `allow-leading-underscore: true`
+- `convention: hyphenatedlowercase`
+
+When enabled, the following are allowed:
+
+```scss
+$foo: hyphenated-lowercase();
```
DanPurdy: As with the tests I would prefer it if we used function declarations in the documentation as that is what most people will be familiar with and expect to see in the documentation. The current examples are just assigning the output of a function to variable and may lead to confusion.

```scss
@function do-something($test) {
  @return $test;
}
```

tests/sass/function-name-format.scss
```diff
@@ -0,0 +1,17 @@
...
+.class {
+  content: snake_case();
+}
+
+@function camelCase($x, $y) {
```
DanPurdy: Could you add a declaration like this for each of the cases please as these are the function declarations that we would be expecting to target before the function is actually used. 
It's also good to check that the function names work for all the cases as they are declared as we've seen a few strange parse errors in other rules with certain definitions.

DanPurdy:
Just a few more notes and then I think this is good to go. Let me know if you don't agree with some of my comments..

tests/sass/function-name-format.scss
benthemonkey: By check, do you mean run the test file through Sass and verify it can be processed correctly?

benthemonkey:
Hows this?

DanPurdy:
:+1:

PR #320, opened 2015-10-17
Update readme with link to make-sass-lint-config

### LINKED ISSUE #299: Add link to scss-lint config converter in README?
I don't mean to self-promote, but I think this tool I've been working on would be useful for folks moving to sass-lint from scss-lint:

http://benthemonkey.github.io/make-sass-lint-config/

Maybe include a link to the tool in the README?

I don't know how this sort of thing works, but would this tool be something valuable to move into the sasstools organization?

benthemonkey:
Closes #299 

Let me know if you have suggestions for a better way to include the link.


bgriffith:
Looks good to me :+1:

PR #323, opened 2015-10-17
Refactor of force nested rules to remove duplication

bgriffith:
PR removes a lot of duplicated functionality from the three force nested rules (`force-pseudo-nesting`, `force-element-nesting` and `force-attribute-nesting`). 

It's decreased the amount of code in each rule by 50% through the creation of 2 new helper functions (`isNestable` and `constructSelector`) and a slight refactor of the rule code.

~~It adds 2 other helper functions (`capitalize` and `camelCaseToHyphens`) to assist with improving the formatting of the warning messages from these three rules.~~
**Update:** 2 string helper functions to improve the formatting of warning messages have been removed and replaced with underscore.string library.


Snugug:
Maybe we consider pulling in underscore.string so we don't need to maintain string manipulation functions on our own. Thoughts?

bgriffith:
Sure makes sense. I'll pull it in and replace the capitalise helper function.

bgriffith:
Ok, by requiring the `underscore.string` library I've removed both the `capitalize` & `camelCaseToHyphens` helper functions.

benthemonkey:
Is there any particular reason you went with Underscore over Lodash? Just curious.

Snugug:
The library is called underscore.string but is not underscore

benthemonkey:
Sorry, thanks for clarifying. But I think my argument still holds. I see that `underscore.string` has functions that [lodash/string](https://lodash.com/docs#camelCase) doesn't, but do you anticipate needing any of those extra functions? If not, I'm trying to suggest that we can exchange dependency on `underscore.string` and instead add specifically `lodash.capitalize`, `lodash.kebabcase`, etc as needed.

DanPurdy:
I agree, I was looking through and we're pulling in a lot of functions in that dependency that we don't need or use. Being able to pull in only the functions we require as a dependency would be favourable over a whole library.

bgriffith:
Completely open on the which library front and happy to pull in individual functions with  `require("underscore.string/capitalize");` etc

DanPurdy:
Individual functions would be better than the whole library each time I would say.

bgriffith:
Done!

benthemonkey:
I don't mean to be too picky here, as this isn't an important area to devote our time, IMO. But to clarify what I meant about Lodash function-specific dependency:

There is a separate NPM module for [each of Lodash's functions](https://www.npmjs.com/browse/keyword/lodash-modularized), so the project itself only depends on the functions within Lodash that you use.

So we'd change `package.json` from having `underscore.string` as a dependency, and instead have [`lodash.capitalize`](https://www.npmjs.com/package/lodash.capitalize) and [`lodash.kebabcase`](https://www.npmjs.com/package/lodash.kebabcase)

If you look carefully at the dependencies of many npm packages, you'll see fewer instances of dependency on Lodash and more on specific Lodash functions.

Just wanted to clarify. No need to spend time changing things.

DanPurdy:
Don't worry @benthemonkey I knew what you meant :wink:

bgriffith:
Thanks for clarifying (and sticking with me 😛) @benthemonkey I see a clear advantage now so happy to go with your approach if everyone's in agreement.

DanPurdy:
Update me!

PR #327, opened 2015-10-18
Add rule: class-name-format

### LINKED ISSUE #324: New Rule: Class Name Format
Add rule to verify name format of class selectors.

Please see the original issue for more information and discussion #17 and #275

benthemonkey:
Slightly different than the other `name-format` rules. It adds `ignore`, an array of whitelisted names.

- [X] Add `allow-leading-underscore` back in
- [x] Add rule to `lib/config/sass-lint.yml`
- [x] Still need to add unit tests for the two helper functions I added.
- [x] More test examples
- [x] Handle suffix extensions to selectors

I welcome feedback on the documentation, as I'm not very familiar with BEM.

Will close #324 

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -139,7 +139,7 @@ helpers.isSnakeCase = function (str) {
 };
 
 helpers.isStrictBEM = function (str) {
-  return /^[a-z]([-]?[a-z0-9]+)*(__[a-z0-9]([-]?[a-z0-9]+)*)?((_[a-z0-9]([-]?[a-z0-9]+)*){2})?$/.test(str);
+  return /^[a-z](\-?[a-z0-9]+)*(__[a-z0-9](\-?[a-z0-9]+)*)?((_[a-z0-9](\-?[a-z0-9]+)*){2})?$/.test(str);
 };
 
 helpers.isHyphenatedBEM = function (str) {
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -331,4 +331,74 @@ helpers.isNestable = function (currentVal, previousVal, elements, nestable) {
   return false;
 };
 
+/**
+ * Tries to traverse the AST, following a specified path
+ * @param   {ASTNode} node           Starting node
+ * @param   {Array}   traversalPath  Array of Node types to traverse, starting from the first element
+ * @returns {Array-of-ASTNode}       Nodes at the end of the path. Empty array if the traversal failed
+ */
+helpers.attemptTraversal = function (node, traversalPath) {
+  var tmp = [],
+      traversalPathRest = traversalPath.slice(1);
+
+  if (traversalPath.length === 0) {
+    return [node];
+  }
+
+  node.forEach(traversalPath[0], function (child) {
+    tmp = tmp.concat(helpers.attemptTraversal(child, traversalPathRest));
+  });
+
+  return tmp;
+};
+
+/**
+ * Collects all suffix extensions for a selector
+ * @param   {ASTNode} ruleset      ASTNode of type ruleset, containing a selector with nested suffix extensions
+ * @param   {string}  selectorType Node type of the selector (e.g. class, id)
+ * @returns {Array}                Array of all selectors (not including prefixed '.', '#' etc) that result from
+ *                                      the suffix extensions
+ */
+helpers.collectSuffixExtensions = function (ruleset, selectorType) {
+  var parentSelectors = helpers.attemptTraversal(ruleset, ['selector', 'simpleSelector', selectorType, 'ident'])
+        .map(function (node) {
+          return node.content;
+        }),
+      childSuffixes = helpers.attemptTraversal(ruleset, ['block', 'ruleset']),
+      selectorList = [];
+
+  if (parentSelectors.length === 0) {
+    return [];
+  }
+
+  var processChildSuffix = function (child, parents) {
+    var currentParents = [],
+        selectors = helpers.attemptTraversal(child, ['selector', 'simpleSelector']),
+        nestedChildSuffixes = helpers.attemptTraversal(child, ['block', 'ruleset']);
+
+    selectors.forEach(function (childSuffixNode) {
+      if (childSuffixNode.length >= 2 &&
+        childSuffixNode.content[0].is('parentSelector') &&
+        childSuffixNode.content[1].is('ident')) {
+
+        parents.forEach(function (parent) {
+          currentParents.push(parent + childSuffixNode.content[1].content);
+        });
+      }
+    });
+
+    selectorList = selectorList.concat(currentParents);
+
+    nestedChildSuffixes.forEach(function (childSuffix) {
+      processChildSuffix(childSuffix, currentParents);
+    });
+  };
+
+  childSuffixes.forEach(function (childSuffix) {
+    processChildSuffix(childSuffix, parentSelectors);
+  });
+
+  return parentSelectors.concat(selectorList);
+};
+
 module.exports = helpers;
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -6,6 +6,11 @@ var util = require('util'),
     yaml = require('js-yaml'),
     merge = require('merge');
 
+/**
+ * Easy access to the 'merge' library's cloning functionality
+ * @param   {object} obj Object to clone
+ * @returns {object}     Clone of obj
+ */
 var clone = function (obj) {
   return merge(true, obj);
 };
@@ -131,22 +136,47 @@ helpers.isLowerCase = function (str) {
   return false;
 };
 
+/**
+ * Determines if a given string adheres to camel-case format
+ * @param   {string}  str String to test
+ * @returns {boolean}     Whether str adheres to camel-case format
+ */
 helpers.isCamelCase = function (str) {
   return /^[a-z][a-zA-Z0-9]*$/.test(str);
 };
 
+/**
+ * Determines if a given string adheres to hyphenated-lowercase format
+ * @param   {string}  str String to test
+ * @returns {boolean}     Whether str adheres to hyphenated-lowercase format
+ */
 helpers.isHyphenatedLowercase = function (str) {
   return !(/[^\-a-z0-9]/.test(str));
 };
 
+/**
+ * Determines if a given string adheres to snake-case format
+ * @param   {string}  str String to test
+ * @returns {boolean}     Whether str adheres to snake-case format
+ */
 helpers.isSnakeCase = function (str) {
   return !(/[^_a-z0-9]/.test(str));
 };
 
+/**
+ * Determines if a given string adheres to strict-BEM format
+ * @param   {string}  str String to test
+ * @returns {boolean}     Whether str adheres to strict-BEM format
+ */
 helpers.isStrictBEM = function (str) {
   return /^[a-z](\-?[a-z0-9]+)*(__[a-z0-9](\-?[a-z0-9]+)*)?((_[a-z0-9](\-?[a-z0-9]+)*){2})?$/.test(str);
 };
 
+/**
+ * Determines if a given string adheres to hyphenated-BEM format
+ * @param   {string}  str String to test
+ * @returns {boolean}     Whether str adheres to hyphenated-BEM format
+ */
 helpers.isHyphenatedBEM = function (str) {
   return !(/[A-Z]|-{3}|_{3}|[^_]_[^_]/.test(str));
 };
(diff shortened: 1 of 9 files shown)


DanPurdy:
I think this needs a lot more test examples, including parent selectors etc, I'll add some examples here later on.

bgriffith:
@benthemonkey Just a heads up that you need to add the rule to `lib/config/sass-lint.yml` for each new rule you create :+1:

Snugug:
And needs to allow a leading underscore. They are valid first characters for class names (and something I personally use)

BPScott:
It looks like the regexs are also missing the escape sequences that [the CSS Spec for identifiers](http://www.w3.org/TR/CSS2/syndata.html#value-def-identifier) defines.

> In CSS, identifiers (including element names, classes, and IDs in selectors) can contain only the characters [a-zA-Z0-9] and ISO 10646 characters U+00A0 and higher, plus the hyphen (-) and the underscore (_); they cannot start with a digit, two hyphens, or a hyphen followed by a digit. Identifiers can also contain escaped characters and any ISO 10646 character as a numeric code (see next item). For instance, the identifier "B&W?" may be written as "B\&W\?" or "B\26 W\3F".

You can take a look at the [CSS Grammar](http://www.w3.org/TR/CSS21/grammar.html#scanner) for  a regex-like interpretation of that rule. Mathias Bynens has a primer on [CSS escapes](https://mathiasbynens.be/notes/css-escapes) with a [test suite of pathologically absurd-but-still-valid class names](https://mathiasbynens.be/demo/crazy-class).


Long story short, [this sort of thing is valid CSS class naming](http://codepen.io/BPScott/pen/QbXBjo).

Snugug:
In a total unrelated note, how many Bens are working on SassLint now? Three Bens, a Dan, and a Sam?

bgriffith:
haha world domination soon :stuck_out_tongue:

benthemonkey:
OK I'll definitely have the class naming adhere to the spec. These regexp's were grabbed straight from scss-lint (as I did with previous rules). But now that you mention it I've wasted hours in the past figuring out why my CSS selector wasn't working, and the problem was an ID started with a number.

However, do you think this rule should perform the rule of class validity checking?

For example, if I had `.101_dalmatians`, should the linter fire two warnings? "Class '.101_dalmatians' should be written in lowercase with hyphens" and "Class names cannot begin with a digit"

DanPurdy:
I would say in the spirit of small single purpose rules we're aiming for we should probably stick with the validity of classnames being handled by a separate rule.

benthemonkey:
I feel like the documentation for this rule is unnecessarily long. Does the user really need to see Good/Bad examples of every convention? Can't I just briefly describe the format of each convention (possibly in a separate documentation page to avoid duplication)? I feel like it would be more concise and consumable by the reader. I would still include examples for the other options.

For instance:
### Convention Examples
- hyphenatedlowercase
  - `header`
  - `header-title`
  - `header-title-active`
- camelcase
  - `header`
  - `headerTitle`
  - `headerTitleActive`
- hyphenatedbem
  - `header`
  - `header__title`
  - `header--inverted`
  - `header--inverted__title`

benthemonkey:
@BPScott I don't think that escape sequences or number codes, while valid, fit within any of the conventions included in this rule. E.g. if a user chooses `hyphenatedlowercase`, I believe they are implicitly allowing only a-z, 0-9 and hyphens.

BPScott:
> In a total unrelated note, how many Bens are working on SassLint now? Three Bens, a Dan, and a Sam?

I refuse to believe my rocking up in the issues and muttering "you missed a bit" counts as working on sass-lint, everyone else contributing code is easily in a different league ;)


@benthemonkey: Good point. The escape code stuff is very much a niche thing and should be avoided 99% of the time. Having this rule for naming conventions with the current regexes, and then a new lower-level rule for `no-invalid-selectors` that deals with the stricter notion of a valid selector - funky escape codes and all - sounds like a good plan.

Then a user can leave the validity check on for a whole codebase, while occasionally explicitly ignoring the naming convention rule for when you DO want a class using escape codes using #70.

DanPurdy:
So for a start with BEM you'll want to check for the following which should all be valid

```scss
.block
.block__element
.block__element--modifier
.block--modifier

.block {
  &__element {

    &--mofifier {

    }
  }
  &--modifier {

  }
}
```

benthemonkey:
@DanPurdy So I tried out your example, utilizing suffix additions to selectors. Gonzales-PE doesn't consider those suffix additions to be part of a class name, and so they are ignored by this linter.

Do you know if I basically have to wait for us to version bump Gonzales-PE so I can utilize the `parentSelectorExtension` node?

DanPurdy:
It should see them as parent selectors no? Looking at your code you're parsing by class, if you were to traverse by selectors as I had to in `no-mergeable-selectors` you'll get much more information. The question then is whether we care about these parent selector additions and whether parsing those is out of scope of actual class name format.

benthemonkey:
Sorry for such a long delay. I played around with accomplishing this, and it was just way too complicated. Here's the location of a class selector, relative to a suffix addition to the selector:

Class selector:
`ruleset -> selector -> simpleSelector -> class -> ident`
Suffix addition
`ruleset -> block -> ruleset -> selector -> simpleSelector -> (parentSelector and ident)`

That's crazy hard to dependably reason about, right? I didn't expect the block of a style definition to be a sibling to the selector, instead of a child of the selector. Not to mention the fact that this has to be recursive, so that multiple nested suffix additions can be processed. Maybe I'm just still getting used to tree traversals. Do you think this can reasonably be tested for?

benthemonkey:
Poking @sasstools/sass-lint-contributors because there has been 10 days without feedback. Are you all ok with not supporting suffix extensions to selectors in the initial version of this rule? If not, could you provide some insight into the best way to robustly accomplish the tree traversal I describe in my previous comment?

tests/yml/.merge-default-rules-false.yml
```diff
@@ -0,0 +1,2 @@
+options:
+  merge-default-rules: false
```
DanPurdy: Any reason for this to be there in this PR?

DanPurdy:
So we're saying it wouldnt support parent selectors?

benthemonkey:
We're going to have to support suffix extensions to selectors when we write a bem-depth rule, so I'll give this another shot. I think I'll need to write some tree-traversal helper functions to accomplish this. Maybe a `searchForSelectorExtensions` function would be nice.

This is a more global issue than I realized. Any selector can be extended so a lot of our rules need fixing.

DanPurdy:
I'm just trying to get at what we're classing as suffix extensions here. If it's all parent selectors then I've already sort of gone through that process with the `no-mergeable-selectors` rule that pretty much constructs every class name correctly though it doesn't format them into a usable format for this rule but you could get close to it.

benthemonkey:
By "suffix extensions to selectors," I was trying to mimic Sass' documentation for the best way to describe adding stuff to the end of an parent's class name. "& must appear at the beginning of a compound selector, but it can be followed by a suffix that will be added to the parent selector" [source](http://sass-lang.com/documentation/file.SASS_REFERENCE.html#parent-selector). For our conversations I thought it would be helpful to have a term to describe this specific use of parent selectors.

I'll look more carefully at `no-mergeable-selectors`.

benthemonkey:
@DanPurdy Quick question. For BEM stuff, once we check suffix extensions to selectors, do we allow stuff like this?

```scss
.block_ {
  &_element-- {
    &modifier {}
  }
}
```

Because the way I'm currently writing this linter, I don't know how I should check for that sort of thing...

gitnik:
While it is valid I would consider it to be anything but good style

DanPurdy:
It would make no sense to do that really and it doesn't follow BEM standards but yes, it's technically valid..

BPScott:
Dropping a vote for "technically valid but certainly not good style so it should raise a warning". Theres all kinds of pathologically stupid things you can do to render the same compiled output, doesn't mean we have to accept all of them. In fact isn't that the whole point of linters? Saying "sure this compiles but you probably shouldn't do that". If somebody is insistent on doing it that way then they can disable the the rule globally or using good old #70.

Snugug:
I like being opinionated, but the goal of our linter should be to let other people be opinionated, not to be opinionated ourselves. If it's valid, we should support it

benthemonkey:
I think I am completely mistaken about that last comment I made. In the example I gave, each of these classes will be individually linted: `.block_`, `.block__element--`, and `.block__element--modifier`. The first two classes will throw lint warnings, and fixing those warnings requires proper use of BEM format.  Let me test this helper method a bit more, update the rule (adding lots of tests), and then I'll tag this PR for further review. Thanks for all the feedback everyone.

benthemonkey:
Alright, ready for review!

Things to note:
- Multiple contiguous underscores are valid for snake_case at the moment (e.g. `snake____case` is valid)
- Because multiple parent selectors can cause a suffix extension to be reported multiple times, using a custom convention message actually reduces the number of warning messages, as there are duplicates.
- Boy oh boy is `helpers.collectSuffixExtensions` complicated to look at. Any suggestions for how to make the function more readable would be appreciated.
- I add another helper function, `helpers.attemptTraversal`, which I think may be generally useful for a number of rules. It takes a node and an array of node types (a path), and returns all nodes which can be reached via that path. I was careful to avoid recursion to improve performance of this function.
- Full code coverage on this rule :smile:

lib/helpers.js
```diff
@@ -3,7 +3,12 @@
 var util = require('util'),
     fs = require('fs'),
     path = require('path'),
-    yaml = require('js-yaml');
+    yaml = require('js-yaml'),
+    merge = require('merge');
+
+var clone = function (obj) {
```
DanPurdy: Can we include JSDoc style comments for each new function in helpers please.

lib/helpers.js
```diff
@@ -131,13 +136,21 @@ helpers.isCamelCase = function (str) {
 };
 
 helpers.isHyphenatedLowercase = function (str) {
-  return !(/[_A-Z]/.test(str));
+  return !(/[^\-a-z0-9]/.test(str));
```
DanPurdy: Missing function comments

lib/helpers.js
```diff
@@ -131,13 +136,21 @@ helpers.isCamelCase = function (str) {
...
 };
 
 helpers.isSnakeCase = function (str) {
   return !(/[^_a-z0-9]/.test(str));
 };
 
+helpers.isStrictBEM = function (str) {
```
DanPurdy: Missing function comments

lib/helpers.js
```diff
@@ -131,13 +136,21 @@ helpers.isCamelCase = function (str) {
...
+  return /^[a-z](\-?[a-z0-9]+)*(__[a-z0-9](\-?[a-z0-9]+)*)?((_[a-z0-9](\-?[a-z0-9]+)*){2})?$/.test(str);
+};
+
+helpers.isHyphenatedBEM = function (str) {
```
DanPurdy: Missing function comments

lib/helpers.js
```diff
@@ -323,4 +336,90 @@ helpers.isNestable = function (currentVal, previousVal, elements, nestable) {
...
 };
 
+/**
+ * Tries to traverse the AST, following a specified path
+ * @param   {ASTNode} node           Starting node
+ * @param   {Array}   traversalPath  Array of Node types to traverse, starting from the first element
+ * @returns {Array}                  Nodes at the end of the path. Empty array if the traversal failed
+ */
+helpers.attemptTraversal = function (node, traversalPath) {
+  var i, nextNodeList, currentNodeList = [],
```
DanPurdy: Split these over multiple lines please

lib/helpers.js
```diff
@@ -323,4 +336,90 @@ helpers.isNestable = function (currentVal, previousVal, elements, nestable) {
...
+      processChildNode = function processChildNode (child) {
```
DanPurdy: Not sure how I feel about child functions in the helpers. Any thoughts?

tests/helpers.js
```diff
@@ -1364,4 +1572,64 @@ describe('helpers', function () {
     assert.equal(expect, result.type);
     done();
   });
+
+  it('collectSuffixExtensions - no extensions', function () {
+    var ruleset = gonzales.parse(['',
+      '.a {\n',
```
DanPurdy: @sasstools/sass-lint-contributors have we got any suggestions here or are we ok with this?

tests/rules/class-name-format.js
```diff
@@ -0,0 +1,231 @@
+'use strict';
+
+var lint = require('./_lint');
+
```
DanPurdy: Can we add the SCSS and Sass block comments as with the other rules here just to give a better visual separation.

https://github.com/sasstools/sass-lint/blob/develop/tests/rules/final-newline.js#L5

tests/sass/class-name-format.scss
```diff
@@ -0,0 +1,96 @@
...
+  @extend .snake_case;
+}
+
+.block-name__elem-name {
+  content: '';
+}
+
+.block {
+  &__element {
+    &--modifier {
```
DanPurdy: Are there any examples where multiple BEM blocks are used? Just to make sure it catches all of them.

i.e.

```
.test
  &__block
    &--mod
  &__block
    &--mod
    &--mod-two
  &--low-mod
```

lib/helpers.js
benthemonkey: It is for this reason that I prefer one var per declaration. Maybe we should write our own eslint rule to check for this.

Snugug: We should write an ESLint rule for this

tests/helpers.js
benthemonkey: I thought this was easier to read than including the already-parsed gonzales AST.

I personally prefer array joins over string concatenation. This can be cleaned up slightly by including the newlines within the `join`

benthemonkey:
@DanPurdy I think I covered all your comments

DanPurdy:
:+1:

PR #329, opened 2015-10-18
Hotfix master: Fix ignored files

### LINKED ISSUE #72: Write tests for CLI
The CLI is currently untested (mostly because I don't know how to write tests for it).

This should be figured out, and tests written

### LINKED ISSUE #129: Ignored file still being processed
`sass-lint -v -i tests/sass/border-zero.scss tests/sass/*.scss`

This results in border-zero.scss being processed still.

DanPurdy:
There was an issue with the CLI where ignored files were still being processed as they weren't being passed through to the file globbing.

I took the opportunity to also add basic tests to the ignored files section of the CLI

I think this should go out as a hotfix in 1.3.2

Closes #72 
Fixes #129 


DanPurdy:
Going to look at a globbing issue as part of this.

DanPurdy:
This is good to go now, 1 thought for @Snugug though if I pass ignored files to the CLI it overwrites the pattern in the users config. `merge-default-rules` documentation states that all other options apart from rules would be merged but this I believe would be counter intuitive when it comes to include and ignore patterns from the CLI if they're specified. Maybe there should be a little extra documentation in the future to make this clear? Or it's quite possible that you believe this behaviour to now be a bug.

@bgriffith I looked into globbing quite heavily, it was working fine but if you didn't wrap a glob pattern in quotes on the CLI then bash was expanding the glob before it got through as an argument to commander. This is the case for the files to be linted and the ignored files. I've updated the documentation to reflect this.

Also @bgriffith if you include a '/' as your first character in a pattern then it will append the pattern to whatever `glob` determines to be the current root. If a root is not specified (which we don't do in sass-lint) then it's the the result of `path.resolve(process.cwd(), "/")` which from the CLI is whatever folder you're currently in or from the grunt/gulp plugins the root of your filesystem usually..

Snugug:
Yah, that’s how I wrote it, under the thinking that anything that was passed in from CLI is implicitly overriding whatever you have (otherwise we’d need a flag for both the option and whether to override that or now)

DanPurdy:
Thought so, wasn't sure it made sense the other way.

Well this is definitely ready to merge then, we should look at the whole CLI and it's workings/tests as part of 2.0.0 also

Snugug:
Agreed. The project has moved too far and too fast for my initial prototype work it's all built around to keep up 😝

DanPurdy:
Ha! I would say it's holding up pretty well! we'll just make 2.0.0 the even more streamlined efficient version!

PR #331, opened 2015-10-18
Update property sort order

### LINKED ISSUE #302: Doesn't handle CSS Modules composition
In [CSS Modules](https://github.com/css-modules/css-modules), you have **compositions**, which basically act like `@extend` in Sass. That's why they should come **first** in the list of properties.

But the `property-sort-order` rule doesn't handle that.

For example:

```
.title
  composes: heading
  background: $background
  box-shadow: 0 2px 5px $shadow
  color: $title
  font-size: 17px
  margin-bottom: 20px
  text-align: center
```

Will output:

```
  14:3  error  Expected `background`, found `composes`    property-sort-order
  15:3  error  Expected `box-shadow`, found `background`  property-sort-order
  16:3  error  Expected `color`, found `box-shadow`       property-sort-order
  17:3  error  Expected `composes`, found `color`         property-sort-order
```

It would be useful if `composes` was ignored when ordering alphabetically.

DanPurdy:
Updates the `property-sort-order` rule to ignore non standard CSS properties to allow custom properties as described in #302 

Closes #302

Looking for possible feedback on this to see if we think this check should be enabled or disabled by an option flag or whether a whitelist would be preferable to this check.

### DIFF
--- a/docs/rules/property-sort-order.md
+++ b/docs/rules/property-sort-order.md
@@ -5,6 +5,7 @@ Rule `property-sort-order` will enforce the order in which declarations are writ
 ## Options
 
 * `order`: `'alphabetical'`, [`'concentric'`](http://rhodesmill.org/brandon/2011/concentric-css/), [`'recess'`](http://twitter.github.io/recess/), [`'smacss'`](http://smacss.com/book/formatting), or `[array of properties]` (defaults to `alphabetical`. Unknown properties are sorted alphabetically)
+* `ignore-custom-properties`: `true`/`false` (defaults to `false`)
 
 Property orders: https://github.com/sasstools/sass-lint/tree/develop/lib/config/property-sort-orders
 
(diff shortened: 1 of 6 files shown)


DanPurdy:
I should also add some extra text to the documentation here.

jgthms:
This looks great :+1: 
Thanks!

DanPurdy:
Going to also add a flag to this to disable this behaviour if needs be. Some might want their 'custom' properties to be ordered.

bgriffith:
Torn on whether default value for `ignore-custom-properties` should be `true` or `false`.

On one hand I feel that custom values could be miss-spelt properties and therefore should not be ignored.

On the other a default action of ignoring anything custom would offer ease of use without either side having to take further actions to get this rule working for them.

DanPurdy:
I think it should remain as false. The default sort order is alphabetical and so it would provide a consistent approach straight out of the box where as it takes a little extra configuration and reading of the documentation to understand why you may not want that.

bgriffith:
Fair point. Let's stick with false.

PR #336, opened 2015-10-20
Hotfix master - Correct no-url-protocols rule name

### LINKED ISSUE #335: Inconsistent rule name (url-format vs no-url-protocols)
The `no-url-protocols` rule has another name in the code:

screenshot 2015-10-20 16 21 46

https://github.com/sasstools/sass-lint/blob/bd5861fcee211591cf8adf9c10495f544a84c26f/lib/rules/no-url-protocols.js

This means that `no-url-protocols` setting in the config is ignored, and `url-format` will be respected. This is not what the docs say.

bgriffith:
Corrects an issue where `no-url-protocols` rule was using an incorrect and now deprecated name.

Fixes #335 

### DIFF
--- a/lib/rules/no-url-protocols.js
+++ b/lib/rules/no-url-protocols.js
@@ -9,7 +9,7 @@ var stripQuotes = function (str) {
 };
 
 module.exports = {
-  'name': 'url-format',
+  'name': 'no-url-protocols',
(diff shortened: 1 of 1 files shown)


DanPurdy:
:+1:

PR #347, opened 2015-10-23
Hotfix Master - no-url-protocols, Fix regex for detecting url protocol use

### LINKED ISSUE #340: no-url-protocols regex too aggressive (reports data-urls)
Here's the offending code (in `/lib/rules/no-url-protocols.js`):

```javascript
var isUrlRegex = /^(http(s?)):\/\/|\/\//;
```

It reports data-urls which have two consecutive slashes in them. The regex should be changed to report urls that begin with `http(s)://` or `//`. There are a number of ways to clean up the regex. Here's my suggested, simplified regex: `/^(https?:)?\/\//`

benthemonkey:
Closes #340 

### DIFF
--- a/lib/rules/no-url-protocols.js
+++ b/lib/rules/no-url-protocols.js
@@ -2,7 +2,7 @@
 
 var helpers = require('../helpers');
 
-var isUrlRegex = /^(http(s?)):\/\/|\/\//;
+var isUrlRegex = /^(https?:)?\/\//;
 
 var stripQuotes = function (str) {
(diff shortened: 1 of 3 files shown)


DanPurdy:
This looks good, just need to update the branch and we can merge it in.

benthemonkey:
rebased.

bgriffith:
We don't generally do it. I think it's easier for everyone to keep track with what's getting merged into develop.

bgriffith:
Update looks good.

Issue #352, opened 2015-10-26
Warnings for nested properties


claudiopro:
This example of nested properties [from the docs](http://sass-lang.com/documentation/file.SASS_REFERENCE.html#nested_properties):

```scss
.funky {
  font: {
    family: fantasy;
    size: 30em;
    weight: bold;
  }
}
```

Produces the following warnings:

```
2:8  warning  Whitespace required before {                         space-before-brace
2:9  warning  Trailing semicolons required                         trailing-semicolon
3:5  warning  Property `family` appears to be spelled incorrectly  no-misspelled-properties
5:5  warning  Property `weight` appears to be spelled incorrectly  no-misspelled-properties
```

claudiopro:
```
$ sass-lint -V
1.3.1
```

DanPurdy:
Confirmed as issue in 1.3.2, we'll look into it, thanks @claudiopro

claudiopro:
@DanPurdy cheers! Let me know how/if I can help

DanPurdy:
@claudiopro 2 of your 3 issues are sorted in 1.3.2 which we just released, the only one still affected here is `no-misspelled-properties` which I was looking into yesterday, still to decide on a good solution for handling nested properties with it but I think it'll either be fixed in 1.3.3 or 1.4.0 which shouldn't be more than a few days / week away.

claudiopro:
That's awesome, thanks for the update!

PR #359, opened 2015-10-27
Hotfix master - Fix trailing-semicolon issue with nested properties

bgriffith:
Fixes the trailing-semicolon false positive highlighted in #352 


tests/sass/trailing-semicolon.scss
```diff
@@ -1,12 +1,20 @@
...
-  content: 'baz', 'fail'
+  padding: 0
 
   .qux {
     content: 'waldo'
   }
 }
+
+.foo {
+  font: {
```
DanPurdy: Can we add some test cases where it should fail here just to make sure?

bgriffith:
Yeah good shout.

DanPurdy:
also try this against it

```scss
.test {
    text: {
        line: {
            through: {
                color: red;
            }
        }
        orientation: landscape;
    }
}
```

bgriffith:
Will do. I think it still needs some work. I'm having a issue.

bgriffith:
Ok added more test cases and also added tests to test for non-default options (include: false)

DanPurdy:
Update me and we're good to go

PR #370, opened 2015-10-29
Fix a small typo in the rule

mitogh:
There was a mistake on this file because it was using the name `indentations`  instead of `indentation` for this rule.


bgriffith:
Hey @mitogh - Thanks for PR, but this is already fixed on develop branch (https://github.com/sasstools/sass-lint/blob/develop/docs/rules/indentation.md) and will be bundled with the next 1.x release :smiley:

mitogh:
Nice! didn't know it, Thank you :+1:

PR #376, opened 2015-11-06
sass syntax mixin include error "Cannot read property 'first' of undefined"

dotdat:
When using Sass syntax mixin includes (+foobar), extends-before-declarations.js expects these to have an atkeyword node and then attempts access properties on an undefined object resulting in the following error. This provides test cases that were able to reproduce the error and a fix to resolve it.

```
/.../node_modules/sass-lint/lib/rules/extends-before-declarations.js:18
          if (item.first('atkeyword').first('ident').content === 'extend') {
                                     ^
TypeError: Cannot read property 'first' of undefined
```

### DIFF
--- a/lib/rules/extends-before-declarations.js
+++ b/lib/rules/extends-before-declarations.js
@@ -13,9 +13,9 @@ module.exports = {
       var lastDeclaration = null;
 
       block.forEach(function (item, j) {
-        if ((item.type === 'include' || item.type === 'extend') &&
-            item.first('atkeyword')) {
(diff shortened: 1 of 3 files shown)


DanPurdy:
Hi @dotdat thanks for this, could you please read the [Contribution guidelines](https://github.com/sasstools/sass-lint/blob/develop/CONTRIBUTING.md) before submitting a PR though.

dotdat:
Sure thing. Sorry for missing that document. I'll close this and start over.

PR #378, opened 2015-11-07
Hotfix Master: negative shorthand values

### LINKED ISSUE #375: shorthand-values rule fails to identify negative values
As the title says, it seems that `shorthand-values` fails to understand this:
```css
test {
    margin: 8px -8px;
}
```

http://puu.sh/lbuOB/8b1c1cb2e0.png

DanPurdy:
Fixes an issue with negative shorthand values, includes extra tests

fixes #375 

DCO 1.1 Signed-off-by: Dan Purdy <danjpurdy@gmail.com>

### DIFF
--- a/lib/rules/shorthand-values.js
+++ b/lib/rules/shorthand-values.js
@@ -83,7 +83,7 @@ module.exports = {
                 });
               }
 
-              else if (val.is('operator') || val.is('ident') || val.is('number')) {
+              else if (val.is('operator') || val.is('ident') || val.is('number') || val.is('unaryOperator')) {
                 fullVal += val.content;
               }
 
--- a/tests/rules/shorthand-values.js
+++ b/tests/rules/shorthand-values.js
@@ -12,7 +12,7 @@ describe('shorthand values - scss', function () {
     lint.test(file, {
       'shorthand-values': 1
     }, function (data) {
-      lint.assert.equal(44, data.warningCount);
+      lint.assert.equal(49, data.warningCount);
       done();
     });
   });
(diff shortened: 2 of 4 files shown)


benthemonkey:
I can't find unaryOperator anywhere in `gonzales-pe`. Where is it?

Besides that, the PR looks good to me.

DanPurdy:
Annoyingly it looks like one of the things that gets removed in `gonzales-pe v 3.0.0` which we still need to wait to update.

See the [changelog](https://github.com/tonyganch/gonzales-pe/blob/3c3c0362a4a668c5f2cb7aaefa6b4cf61a6d46f8/CHANGELOG.md#node-types-2)

I think that's why this is probably missing in the first place as this rule was written around the time gonzales was updating I believe so unaryOperator would of been missing from the docs. Still it's my own fault for not adding negative values to the tests in the first place.

benthemonkey:
So do you think my feedback is sufficient to merge? Or do you want to wait for another contributor?

DanPurdy:
I'm confident that this fixes the issue reported and the tests should prove that also.

If you'd prefer to wait for someone else to give it the once over too then that's ok, there's no immediate rush to get `1.3.3` out as most of us have been incredibly busy recently it seems, on the other hand if you're happy that the tests cover the rule sufficiently and the code added is formatted correctly and indeed fixes the issue reported then feel free to merge.

In response to your feedback, once we update to the latest gonzales there's going to be a little bit of rewriting going on anyway to fix some of the rules broken by the latest gonzales changes. The fact that this node type won't exist anymore won't actually break the rule though as it'll just always return `false`.

:smile:

benthemonkey:
Sounds good. So I'd like to merge both PRs. Just so I don't accidentally mess anything up, can you walk me through the steps here? I'd also be interested to know the specific steps you follow for releases as well. I know about publishing to npm, creating releases, etc, but I want to follow the same steps as you.

alfaproject:
LGTM

bgriffith:
Looks good :shipit:

Issue #371, opened 2015-10-30
New rule: Property units


bgriffith:
One rule that we are missing is one to lint which units can be used with specific property values.

SCSS Lint currently does this with https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#propertyunits though I'm undecided if this is the best way, or if we should consider breaking it up into multiple unit managing rules.

Open to thoughts

DanPurdy:
I was looking at this the other day, I think it would be a mistake to split this into a per property rule, i.e. a separate rule for border units, margin units etc, there would be so much duplication and complication for users and would make it difficult to implement global units.

bgriffith:
Closed with #387

PR #387, opened 2015-11-10
Add feature to allow only certain units to be used in a property

donabrams:
As per https://github.com/sasstools/sass-lint/issues/371, I migrated https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#propertyunits over. 

DCO 1.1 Signed-off-by: Donald Abrams <donabrams@gmail.com>


Snugug:
A couple of comments, but otherwise looks good

benthemonkey:
This looks great so far! Here are my initial comments:
- Can you add some additional unit types in the test files? It's always good to include as many cases as possible.
- Before this gets merged you'll need to add a documentation file to `/docs/rules` and add this rule to `/lib/config/sass-lint.yml`

Snugug:
Good catches @benthemonkey

lib/rules/property-units.js
```diff
@@ -0,0 +1,54 @@
...
+
+module.exports = {
+  'name': 'property-units',
+  'defaults': {
+    'unitsAllowedForProperties': {},
+    'globallyAllowedUnits': []
+  },
+  'detect': function (ast, parser) {
+    var result = [];
+    var globallyAllowedUnits = parser.options.globallyAllowedUnits;
```
DanPurdy: Mixin var declaration styles here and [here](https://github.com/sasstools/sass-lint/pull/387/files#diff-ecba41c08d8d6d5f4222fcbe86d47722R18), if we stick with just trailing commas in this instance.

donabrams:
Integrating feedback now.

lib/rules/property-units.js
```diff
@@ -0,0 +1,54 @@
...
+    var unitsAllowedForProperties = parser.options.unitsAllowedForProperties;
+
+    ast.traverseByType('declaration', function (declaration) {
+      var declarationType = declaration.content[0].content[0].type,
+          propertyName = declaration.content[0].content[0].content,
```
DanPurdy: This is fairly dangerous, we should be checking each content block exists and is a type we are expecting. If any of these are undefined for whatever reason this will throw a fatal error. 

It's best here to use the ast's forEach method over the content blocks checking for the type you're looking for to make sure, at least that way we can fail gracefully and possibly have false positives rather than fatal errors.

lib/rules/property-units.js
```diff
@@ -0,0 +1,54 @@
...
+          valueNode = declaration.content
+            .filter(function (node) {
+              return node.type === 'value';
+            })[0];
+      if (declarationType === 'ident' && valueNode.content[0].type === 'dimension') {
+        var dimensionNode = valueNode.content[0];
+        var unitType = dimensionNode.content[1].content;
```
DanPurdy: Same as above, multiple var declarations instead of trailing commas.

lib/rules/property-units.js
```diff
@@ -0,0 +1,54 @@
...
+        var unitsAllowedForProperty = unitsAllowedForProperties[propertyName];
+        // If a property is defined in unitsAllowedForProperty, then it will only validate those unit types
+        if (unitsAllowedForProperty) {
+          if (unitsAllowedForProperty.indexOf(unitType) === -1) {
+            result = helpers.addUnique(result, {
+              'ruleId': parser.rule.name,
+              'severity': parser.severity,
+              'line': dimensionNode.start.line,
+              'column': dimensionNode.start.column,
+              'message': 'values for property \'' + propertyName + '\' may not use ' + unitType + ' units'
```
DanPurdy: Can we capitalize the first word of the warning message please.

lib/rules/property-units.js
```diff
@@ -0,0 +1,54 @@
...
+        }
+        // If no units are defined in globallyAllowedUnits, then allow all of them
+        // Otherwise, verify the given unit is in the globallyAllowedUnits list.
+        else if (globallyAllowedUnits.length && globallyAllowedUnits.indexOf(unitType) === -1) {
+          result = helpers.addUnique(result, {
+            'ruleId': parser.rule.name,
+            'severity': parser.severity,
+            'line': dimensionNode.start.line,
+            'column': dimensionNode.start.column,
+            'message': 'values for property \'' + propertyName + '\' may not use ' + unitType + ' units'
```
DanPurdy: Capitalization again here please.

lib/rules/property-units.js
benthemonkey: Shouldn't this be in our .eslintrc?

DanPurdy: I don't think this is something we've been enforcing but within rules it makes sense to stick with one style.

tests/sass/property-units.sass
```diff
@@ -0,0 +1,4 @@
+.blue
+  height: 3em
+  width: 2px
+  border-top: 2px
```
DanPurdy: I would also second @benthemonkey here and add a lot more test cases, including multiple statement types, variable declarations, unit-less values and long form short form values too.

tests/rules/property-units.js
```diff
@@ -0,0 +1,149 @@
...
+      done();
+    });
+  });
+
+  it('globally-allowed-units: [\'px\'], units-allowed-for-properties: {}', function (done) {
+    lint.test(file, {
+      'property-units': [
+        1,
+        {
+          'globally-allowed-units': ['px']
```
DanPurdy: Could we also add examples here of allowing multiple unit types.

DanPurdy:
@donabrams added a few comments, other than those this is looking good, will give it another look over once you've added in the changes but really appreciate you taking the time to work on this.

donabrams:
OK, I think I address all the comments. I'm not thrilled about allowing variables of the wrong type in (see docs), but after playing with it for a while (see https://github.com/donabrams/sass-lint/commit/4e32bdc9e4d028450884b2bad561a37512ee7df5), it's the best I can come up with that isn't full of rather terrible surprises. Any ideas?

Snugug:
@donabrams looks like you didn't catch @DanPurdy's reply to my comment. Can you change the options to `global-units` and `property-units`?

Also, for the documentation, please document `global-units` and `property-units` separately as well as together. Please also be more clear in the description of `global-units` and `property-units` what those options are controlling and, especially for `property-units`, what's expected (key/value, what's expected, there, etc…).

Finally, I may have missed it, but I don't see the rule added to our [master list of rules](https://github.com/sasstools/sass-lint/blob/develop/lib/config/sass-lint.yml)

Snugug:
Actually, as the rule is called `property-units`, can the options be `global` and `by-property`? @DanPurdy @benthemonkey thoughts on that before @donabrams makes that change?

Snugug:
And while I'm stream-of-thoughting, in the documentation, when describing the property specific option, be sure to indicate there that it will override the global option for that given property just so there is no confusion.

Snugug:
Please, and thanks for your work on this!

DanPurdy:
`individual-property-units` ?

benthemonkey:
:+1: for `per-property`

donabrams:
All right, does this look good to everyone?

benthemonkey:
LGTM!

bgriffith:
Late to the party! Looking good @donabrams. Good to see this rule is sorted :+1:

bgriffith:
:shipit:

PR #395, opened 2015-11-12
Fix false positive for valid css with space-around-operator

### LINKED ISSUE #394: Space around operator false positive
The space-around-operator rule recently added to dev branch has an issue with valid css using negative values.

```css
.foo { margin: -15px -15px; }
``` 

Sass 1.3.2, node 5.0.0

bgriffith:
This PR fixes a bug where valid css would: 
1 - Be flagged up as requiring a space
2 - Cause a error with the parser.

The css in question is negative values for margin, top etc

`margin: -15px;`
 
Ps. Thanks to @DanPurdy for catching it!

Closes #394 


DanPurdy:
How about variable/selector names, I know they're not targeted but in anticipation of gonzales changes that may affect this it would be good to have them in?

bgriffith:
I don't feel that's necessary. Since we're only ever targeting `['condition', 'atruleb', 'value', 'atrulerq']` to begin with, selectors should never come up. I know you could potentially future proof it (somehow?!) but I can't see any reason why a character within a selector would ever be classed as an operator.

DanPurdy:
'should' :wink: 

Update this and it's good to go

DanPurdy:
grr update again and let me know.

bgriffith:
I've clicked update. Now it's just waiting game...

PR #399, opened 2015-11-12
Hotfix Master - add coveralls

benthemonkey:
Sorry! I broke merging into master because of coveralls.

I cherry-picked the two relevant commits from develop.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ node_js:
   - node
   - iojs
 before_script: npm link
+after_success: npm run coveralls
 notifications:
   slack:
     secure: RrEbq2xE1hWdog4AckkaKDnIYYwo5VdjPcFNhRJbn/7KI0fKeZVCKZy1Ww7aaJGth7R7UX415sEV1U6RrjFyhnBb6Sh+rh8fKTvcvuTbENZW45SbtUD+xmgOvb2kfk4PzgD5Q457DpchAZD7W+E+9qr3xI3Uvh4II1uhDmSKiLI=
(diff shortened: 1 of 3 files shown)


benthemonkey:
Whyyyyyy

benthemonkey:
@Snugug Do you have coveralls configured to have some coverage threshold? I don't see why the build would fail without a specific coverage requirement being set.

DanPurdy:
Ha it's set to 95% I believe seen as develop was so high in coverage. I think once 1.4.0 is out we should set the limit at 95%, until then it's probably best to just remove this limit for now..

benthemonkey:
So is the setting changed? Should I force push to get it to re-run?

Snugug:
You won't be able to force push. I'll disable it momentarily.

Snugug:
Okay, updated and re-run

benthemonkey:
I've been force-pushing on my forked branches without any issues for some time now. It looks like coveralls isn't reporting back to GitHub now... I'm so sorry about the amount of your time this is taking.

benthemonkey:
Here is the error from Travis: https://travis-ci.org/sasstools/sass-lint/jobs/90841691#L817

Snugug:
Ahh, to yours, not master or develop. 

When I posted it looked like it passed

benthemonkey:
Alright! Looks like we're passing. @DanPurdy or @Snugug want to merge?

DanPurdy:
:+1: I'm still not convinced by this coveralls though, the counts seem a little off and the fact it keeps not reporting a status..

PR #401, opened 2015-11-13
Add cacheable config

### LINKED ISSUE #279: Cache contents of config file
Placing a console.log inside config.js shows that we're loading the config from scratch once per rule. We should be loading the config once per invocation of sass-lint.

I'm referring to using sass-lint from the CLI, and specifying a config file. e.g. `sass-lint -c .sass-lint.yml`

DanPurdy:
By default the config is now cached on each invocation of sass-lint with an option to override this behaviour. 

You can override this behaviour if you want, an example would be someone tweaking their `sass-lint` config within a IDE plugin.

Updated the tests to work with this new implementation.

Some of the crude benchmarks I ran for loading the config when cached showed an average of a 26% speed increase loading each rule as opposed to always loading from scratch though i'm sure with tweaks we could improve this and definitely an opportunity with 2.0

Looking for feedback here about if everyone is happy for this to move forward. It follows our usual pattern of preference i.e. `options > user config > defaults`

closes #279 

`DCO 1.1 Signed-off-by: Dan Purdy <danjpurdy@gmail.com>`


DanPurdy:
Also closes #403

DanPurdy:
Ready for review now

benthemonkey:
A little bit off-topic, but do we need the devDependency on the `deep-equal` module? My understanding is the reason to use it over `assert.deepEqual` is to return a boolean instead of throwing an error. However, we only use it in tests, where throwing is fine.

benthemonkey:
Other than that, LGTM

DanPurdy:
Thanks @benthemonkey, I've tried to cover every eventuality with the tests and they're actually doing something now so good catch there! @bgriffith and @Snugug should probably look this over before we merge it as it's quite far reaching in what it does.

As for deep equal, correct me if i'm wrong but we'd need to use the `assert.deepStrictEqual` method which I 'believe' isn't supported in earlier versions of Node that we test against still? I don't mind removing if we need to though.

benthemonkey:
Oops, you're right, 0.10.x doesn't support `assert.deepStrictEqual`.

Snugug:
In general I like this rule. Think it should be off by default as it's one of those things that can potentially be a gotcha bug that's hard to track down.

DanPurdy:
Ok, i'll update this so it isn't enabled by default then, I think that does make sense.

DanPurdy:
Ok updated the defaults to false now. Ready for final review.

benthemonkey:
This LGTM (and someday that'll mean something!)

DanPurdy:
Feel free to merge it, I just wanted the other guys to give it a once over before we merged it in the first place. All good now though.

tests/config.js
```diff
@@ -1,102 +1,192 @@
...
+      'ignore': [
+        'foo',
+        'bar'
+      ]
+    },
+    'rules': {
+      'no-duplicate-properties': 0,
+      'indentation': [
+        2,
+          {
```
bgriffith: Funky indentation?

DanPurdy:
Turns out the indentation was fine @bgriffith, though it looks wrong to me!

bgriffith:
Sorry about that! :shipit:

Issue #383, opened 2015-11-10
@at-root registers as nested depth


gitnik:
```scss
.foo {
    @at-root &__bar {
        @at-root &__baz {
            display: inline-block;
        }
    }
}
```

This registers as a depth of 2 which is technically correct but considering that `@at-root` generates a completely new, unnested selector I think this rule should be disabled in this specific case. I am willing to submit a PR if this behaviour is not wanted anymore.

DanPurdy:
Hi @gitnik I think this makes sense yes, my only worry is that it opens a lot of edge cases around the rule that we'll need to be very careful with. @sasstools/sass-lint-contributors does anyone have any other opinions on this regarding whether we are really linting compiled nest depth or actual nesting depth within the scss/sass files where I would still class the above as nesting depth of 3.

gitnik:
Well I don't really consider this "nesting" since the end result will be everything but nested. But I can see where you're coming from. Looking forward to what other people have to say about this

DanPurdy:
I had another think about this @gitnik I think you're right it should be excluded. Happy for you to get a PR up if you want to.

bgriffith:
Are we saying that the following should be valid? 

```scss
.foo {
@at-root &__bar {
@at-root &__baz {
    display: inline-block;
}
}
}
```

gitnik:
IMO, yes

benthemonkey:
The use of `@at-root` is extraneous in this example. Suffix extensions are not nested within their parents (so removing `@at-root` outputs the same CSS, [example gist](https://gist.github.com/benthemonkey/b089848bce1833381598)). So if we are to be consistent, we need to treat the use of `@at-root` and suffix extensions identically.

brunowego:
Take a look into my case:

```bash
./src/styles/_form.scss
  46:1  error  Please check validity of the block starting from line #46  Fatal
```

```scss
.form-input {

    @at-root div#{&} {

    }

    @-moz-document url-prefix() {
        @at-root div#{&} {

        }
    }

    @at-root textarea#{&} {

    }
}
```

DanPurdy:
@brunowego this again is the AST you need to report this there

https://github.com/tonyganch/gonzales-pe/issues

We don't have any control over these parse issues unfortunately

brunowego:
Thanks @DanPurdy :+1:

PR #404, opened 2015-11-14
Fix depth lint

gitnik:
Only report nesting depth if generated CSS actually is nested

### DIFF
--- a/lib/rules/nesting-depth.js
+++ b/lib/rules/nesting-depth.js
@@ -2,11 +2,9 @@
 
 var helpers = require('../helpers');
 
-var lodash = require('lodash');
-
 var compiledCSSDoesNotCreateDepth = function (selector) {
-  return lodash.isArray(selector.content) &&
-    lodash.isArray(selector.content[0].content) &&
+  return Array.isArray(selector.content) &&
+    Array.isArray(selector.content[0].content) &&
     selector.content[0].content[0] === '&';
 
 };
--- a/lib/rules/nesting-depth.js
+++ b/lib/rules/nesting-depth.js
@@ -2,6 +2,15 @@
 
 var helpers = require('../helpers');
 
+var lodash = require('lodash');
+
+var compiledCSSDoesNotCreateDepth = function (selector) {
+  return lodash.isArray(selector.content) &&
+    lodash.isArray(selector.content[0].content) &&
+    selector.content[0].content[0] === '&';
+
+};
+
 module.exports = {
   'name': 'nesting-depth',
   'defaults': {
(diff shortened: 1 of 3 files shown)


gitnik:
Fixes https://github.com/sasstools/sass-lint/issues/383

DanPurdy:
Thanks @gitnik Can you please include a DCO with this [PR](https://github.com/sasstools/sass-lint/blob/develop/CONTRIBUTING.md#pull-requests)

lib/rules/nesting-depth.js
```diff
@@ -2,6 +2,15 @@
 
 var helpers = require('../helpers');
 
+var lodash = require('lodash');
+
+var compiledCSSDoesNotCreateDepth = function (selector) {
+  return lodash.isArray(selector.content) &&
```
DanPurdy: We don't have this package available, we were using certain modules of loadash rather than including the whole library.

gitnik: Ah I already feared. Shall I add an `isArray` method to the helpers then?

gitnik:
`DCO 1.1 Signed-off-by: Nik Karbaum <nik.karbaum@gmail.com>`

lib/rules/nesting-depth.js
DanPurdy: You could just include the lodash.isarray package if you require it, no point writing and testing our own.

gitnik: Ok I'm just using the native `isArray` method now since I can always expect that to be available in a node environment

benthemonkey:
According to Coveralls, none of the current unit tests actually trigger the check you added about 0 depth.

https://coveralls.io/builds/4149812/source?filename=lib%2Frules%2Fnesting-depth.js#L32

gitnik:
Urgh of course it doesn't work, sorry. Github shouldn't allow commits after 12 PM. I'll get back to this tomorrow

DanPurdy:
@gitnik Looking good, just a few untested lines to cater for if you get the chance.

https://coveralls.io/builds/4149953/source?filename=lib%2Frules%2Fnesting-depth.js

lib/rules/nesting-depth.js
```diff
@@ -2,6 +2,13 @@
 
 var helpers = require('../helpers');
 
+var compiledCSSDoesNotCreateDepth = function (selector) {
```
DanPurdy: Can we stick some jsDoc style comments in for this function?

gitnik:
I will take care of all the suggested improvements over the weekend :-)

DanPurdy:
No problem, thanks!

DanPurdy:
I'm going to remove this from the `1.4.0` release for now, it's not fixing the issue raised and will need further work with regards to `at-rules` in a similar fashion to what can be found with the property-count rule in regards to how they're treated in `gonzales`.

Will look into it.

PR #407, opened 2015-11-15
Feature/property count

### LINKED ISSUE #406: New Rule: Property Count
Add rule property count as per [scss-lint](https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#propertycount)

DanPurdy:
Adds the rule `property-count` which enforces the maximum number of properties per ruleset. This can be configured to be per ruleset or inclusive of nested rulesets.

Warning Message:
> Rule set contains {number} {property/properties} more than the specified maximum of {number} {(includes properties in nested rules)}

Includes the options:
* `max-properties`
* `include-nested`

Closes #406

### DIFF
--- a/docs/rules/property-count.md
+++ b/docs/rules/property-count.md
@@ -1,0 +1,122 @@
+# Property Count
+
+Rule `property-count` will enforce the maximum number of properties allowed within a ruleset. This can include or exclude a cumulative count of nested properties also.
+
+## Options
+
+* `max-properties`: number (defaults to `0`)
+* `include-nested`: `true`/`false` (defaults to `false`)
+
+## Examples
+
+When `max-properties: '2'`, the following are allowed.
+
+```scss
+.foo {
+  content: '';
+  display: block;
+}
+
+.nest {
+  content: '';
+  display: block;
+
+  .nest-child {
+    content: '';
+    display: block;
+  }
+}
+```
+When `max-properties: '2'`, the following are disallowed.
+
+```scss
+.foo { // lint reported
+  content: '';
+  display: block;
+  color: red;
+}
+
+.nest { // lint reported
+  content: '';
+  display: block;
+  color: red;
+
+  .nest-child { // lint reported
+    content: '';
+    display: block;
+    color: red;
+  }
+}
+```
+
+When `max-properties: 2` and `include-nested: true`, the following are allowed.
+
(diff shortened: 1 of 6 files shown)


benthemonkey:
I see that any node type that interrupts the `block -> ruleset -> block ...` cycle stops the code from checking for further-nested properties. This means that properties nested inside media queries, `@at-root`, etc. are not included in the nested property count. Is this the desired behavior?

DanPurdy:
With at-root definitely because it's about making sure you're using the cascade correctly and at-root removes that consideration anyway. As for media queries, that's a question really, do we want to handle that?

I ran this implementation as well as `scss-lint`s version and found similar issues with media queries and at-root so id be open to seeing other scenarios we should support.

benthemonkey:
Additionally, until we upgrade Gozales, suffix extensions are being included in nested property counts, even though they aren't nested within their parent when compiled. However, suffix extensions are nested within their grandparent. Perhaps this exception doesn't fall within the "spirit" of this rule. What do you think? Example:

```scss
.foo {
  .bar {
    . color: green;
    &-baz {
      color: blue;
    }
  }
}
```
With `include-nested: true`, `.bar`'s property count ought to be 1, while either `.foo`'s or `.foo .bar-baz`'s property count is 2.

DanPurdy:
It's not about how many properties are on the rule after compilation though its about how many properties are within each block in the source. When the nested option is on with the above code the nested count should always be 2 and the unnested counts should be 0, 1, 1

DanPurdy:
Will update this to work with mixins etc

DanPurdy:
Gonzales scuppers the `.sass` version of this when using `at-root` or media queries as it considers atrules to be siblings of the properties they contain rather than parents which it does in `.scss`... :cry:

Snugug:
Ugh. GONZALES!

Yah, I'm OK with that for the moment as long as that's documented.

Maybe we should start adding what syntaxes are supported for each rule if it differs from "all"

DanPurdy:
I'll be moving forward with this for the latest version of gonzales-pe as I believe the issue I was encountering with the Sass syntax should be fixed!

DanPurdy:
I'll be reviving this for the new 1.6 update now once I've verified that sass blocks are indented correctly

PR #418, opened 2015-11-19
New Rule: bem-depth

### LINKED ISSUE #413: New Rule: BEM Depth
Name: `bem-depth`
Option: `max-depth`, defaults to 1

Add a rule that enforces a maximum number of elements in a BEM selector.

Note that the option name, `max-depth` is different from scss-lint, which calls the option `max_elements`. This way the option is consistent with nesting-depth, and soon selector-depth).

The helper function I wrote, `collectSuffixExtensions`, should make this fairly straightforward to implement.

benthemonkey:
Pretty straightforward.

Will close #413 

### DIFF
--- a/lib/rules/bem-depth.js
+++ b/lib/rules/bem-depth.js
@@ -1,0 +1,72 @@
+'use strict';
+
+var helpers = require('../helpers');
+
+/**
+ * Get number of BEM elements in
+ * @param   {string}  str String representing a class selector
+ * @returns {integer}     Number of BEM elements in str
+ */
+var bemDepth = function (str) {
+  var elements = str.split('__').length;
+
+  if (elements >= 2) {
+    return elements - 1;
+  }
+
+  return 0;
+};
+
+module.exports = {
+  'name': 'bem-depth',
+  'defaults': {
+    'max-depth': 1
+  },
+  'detect': function (ast, parser) {
+    var result = [];
+
+    ast.traverseByTypes(['ruleset', 'placeholder'], function (node) {
+      var name,
+          depth,
+          selectorAndExtensions,
+          maxDepth = parser.options['max-depth'];
+
+      if (node.is('placeholder')) {
+        name = node.first('ident') && node.first('ident').content;
+        depth = bemDepth(name);
+
+        if (name && depth > maxDepth) {
+          result = helpers.addUnique(result, {
+            'ruleId': parser.rule.name,
+            'line': node.start.line,
+            'column': node.start.column,
+            'message': ['Placeholder \'%', name, '\' should have ', maxDepth, ' or fewer BEM elements, but ',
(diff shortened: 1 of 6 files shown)


lib/rules/bem-depth.js
```diff
@@ -0,0 +1,73 @@
...
+ * @param   {string}  str String representing a class selector
+ * @returns {integer}     Number of BEM elements in str
+ */
+var bemDepth = function (str) {
+  var elements = str.split('__').length;
+
+  if (elements >= 2) {
+    return elements - 1;
+  }
+  else {
```
DanPurdy: Unnecessary else here. Just return 0 after the if.

tests/rules/bem-depth.js
```diff
@@ -0,0 +1,93 @@
...
+//////////////////////////////
+// SCSS syntax tests
+//////////////////////////////
+describe('bem depth - scss', function () {
+  var file = lint.file('bem-depth.scss');
+
+  // silly case to raise a warning for every selector/placeholder
+  it('[max-depth: 0]', function (done) {
+    lint.test(file, {
+      'bem-depth': [1, { 'max-depth': 0 }]
```
DanPurdy: max depth 0 should probably disable the rule rather than generate errors/warnings. This would also prevent people using negative values?

DanPurdy:
Few little comments, and possibly we should include some test examples where there are mutliple same level elements again just to make sure we aren't getting any false positives. Though I know these are being tested in classname format as you're using the same functions.

Apart from that, it'll be great to have this rule! Looks good to me :+1:

tests/rules/bem-depth.js
benthemonkey: I think that the value of an option shouldn't change the severity of the rule set by the user. E.g.
```
bem-depth:
  - 2
  - max-depth: 0
```
shouldn't silently turn off the rule. How about, instead:
- Specify max-depth > 0 in the docs
- Error if max-depth <= 0 inside this rule

DanPurdy: Hmm on second thoughts allow 0 to work I think but error on negative values. If someone sets it to 0 the warning message provided should provide enough information to see their error.

benthemonkey: Out of curiosity I checked out ESLint and it will let you set the indentation to negative numbers, if you want to. I think this is a case where we let the user be dumb if they want.

DanPurdy: Ok agreed. Let's just handle the unnecessary else above then and this will be good to go.

benthemonkey:
@DanPurdy done. The changes were small enough that i just squashed them into the original commit.

DanPurdy:
Good to go once it's updated..

benthemonkey:
@DanPurdy updated.

Issue #414, opened 2015-11-19
Allow empty entries in property-sort-order declaration indicating visual separation of property blocks


gajus:
From scss-lint [`PropertySortOrder`](https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#propertysortorder):

> You can enforce that "groups" of properties be visually separated by setting the separate_groups option to true. When specifying a custom order, you can indicate that you want two groups of properties to be visually separate by inserting an empty item, e.g.

```yml
linters:
  PropertySortOrder:
    order:
      - display
      - position
      -            # This empty element signals a visual separation
      - margin
      - padding
    separate_groups: true
```

which requires that CSS properties are grouped, e.g.

```css
display: block;

maring: 0;
```

It would be great if [`property-sort-order`](https://github.com/sasstools/sass-lint/blob/develop/docs/rules/property-sort-order.md) supported this. If it does already, it is not documented anywhere.

DanPurdy:
Thanks for this.

There was discussion about this already i believe in #350

gajus:
@DanPurdy I am writing a PR for this.

There is one condition that I cannot understand:

https://github.com/sasstools/sass-lint/blob/728c07fb5cc609bf5854605c06eeebb26a16142e/lib/rules/property-sort-order.js#L35-L72

First you check if `order` is string. Then you check if `order` is an object.

The `else {` block will never be reached? (what else can `order` be?)

gajus:
Writing expectations in such a way makes 0 sense.

https://github.com/sasstools/sass-lint/blob/develop/tests/rules/property-sort-order.js#L249

When a single test breaks, all the other tests break too. And you have zero clue as to what is the expectation of the error.

gajus:
Either my or your implementation of `property-sort-order` is wrong, because all of the tests are failing.

I have isolated an example.

Here is the subject of the test:

```css
.other-property {
  composes: heading;
  height: 100vh;
  display: block;
  width: 100vw;
  border: 1px;
}
```

Here is the error that my script produces:

```json
{
    "filePath": "property-sort-order.scss",
    "warningCount": 2,
    "errorCount": 0,
    "messages": [
        {
            "ruleId": "property-sort-order",
            "line": 4,
            "column": 3,
            "message": "Expected `width`, found `display`",
            "severity": 1
        },
        {
            "ruleId": "property-sort-order",
            "line": 5,
            "column": 3,
            "message": "Expected `display`, found `width`",
[...]

Issue #350, opened 2015-10-23
property-sort-order: allow sorting by groups of rules


mknadler:
:wave: 

The use case I am envisioning here is something like:

given a custom order of properties such as:

```
- 'color'
- 'font'
- 'font-family'

- 'position'
- 'top'
- 'left'
```

It would be great if there were a way to signify that I don't care about the order _inside_ of a chunk of properties, just _amongst_ a chunk of properties -- i.e., that I don't care if `color` comes before `font-family`, but both should come before `position`.

bgriffith:
Hey @mknadler - I agree. It would be really useful to be able to order groups instead of individual properties. 

Open to anyone's thoughts on how we could approach this.

track0x1:
Have each grouping in its own array and make sure every property is always less than or equal to the current array's index?

DanPurdy:
Yeah that's the simple way it's just how manageable we want to make it. The current rule enforces alphabetical ordering for any unspecified properties in a property order so we'd need to probably revert that. I'm just looking into property sort order as part of #331 so I can get an idea of what the best approach would be from that.

DanPurdy:
Is this how the [concentric](https://github.com/sasstools/sass-lint/blob/develop/lib/config/property-sort-orders/concentric.yml) sort order should be working out of interest?

occupant:
@DanPurdy The full, up to date list for Concentric is here:
https://github.com/brandon-rhodes/Concentric-CSS/blob/master/style3.css

The one big omission I see right away is box-sizing. That should be first. Right now it gets added after everything else. Of course, it's critical to determining the correct box model, and -- since that's the way concentric orders properties -- it's kind of a big deal.

Also missing are CSS grid specific properties, which I'd imagine would be ordered something like this:
```
grid
grid-area
grid-template
grid-template-areas
grid-template-rows
grid-template-columns
grid-column
grid-column-start
grid-column-end
grid-row
grid-row-start
grid-row-end
grid-auto-rows
grid-auto-columns
grid-auto-flow
grid-gap
grid-row-gap 
grid-column-gap
```

I'll make a pull request for these updates.

totkeks:
I just came across this issue when looking for a way to group scss properties after finding this documentation: https://github.com/sasstools/sass-lint/blob/master/docs/rules/property-sort-order.md

@occupant any update on your pull request?

occupant:
[...]

Issue #139, opened 2015-09-08
Pluggable Rules


Snugug:
We should provide a way for end users to define and distribute rules as NPM packages that will be automatically picked up and be usable by Sass Lint

- [ ] Look to ESLint as model
- [ ] Recurse down dependency tree to find rules
- [ ] Create naming convention

gajus:
I am happy to take responsibility of this, assuming everyone agrees with the proposed naming convention and implementation logic, https://github.com/sasstools/sass-lint/issues/432.

gajus:
Though, I am not happy that this has a milestone v2.0.0. This is not a breaking change.

Snugug:
While _this_ is not a breaking change, it's part of the suite of things we're going to be releasing in 2.0, which will have enough breaking changes that it will effect anything that touches the AST. We don't want to introduce something like this for a point version and then quickly deprecate all of that work for the changes we're introducing.

glen-84:
Any news here? Is there no other way to add custom rules at the moment?

kmiyashiro:
>We don't want to introduce something like this for a point version and then quickly deprecate all of that work for the changes we're introducing.

I wouldn't consider 1.4+ years years quickly deprecating something 😅 . Unless 2.0 is imminent, maybe pluggable rules can be considered as a 1.x feature? Is there a way to define custom rules without creating separate modules currently?

PR #423, opened 2015-11-20
Updated sorting logic and error reporting. Added (group separator).

gajus:
This step is required to implement https://github.com/sasstools/sass-lint/issues/414.

The most notable, user affecting change is the error formatting for `property-sort-order` rule.

The new format is:

```js
  'ruleId': parser.rule.name,
  'line': block.start.line,
  'column': block.start.column,
  'message': 'Expected "' + expectedOrder + '", found "' + currentOrder + '"',
  'severity': parser.severity
});
```

and it is reported per block, as opposed to per property.

### DIFF
--- a/lib/rules/property-sort-order.js
+++ b/lib/rules/property-sort-order.js
@@ -5,16 +5,23 @@ var helpers,
     fs,
     path,
     sortBy,
+    clone,
     propertyCheckList,
     orderPresets,
     getOrderConfig,
-    sortProperties;
+    sortProperties,
+    trimGroupSeparator,
+    validateSeparation,
+    GROUP_SEPARATOR;
+
+GROUP_SEPARATOR = '(group separator)';
 
 helpers = require('../helpers');
 yaml = require('js-yaml');
 fs = require('fs');
 path = require('path');
 sortBy = require('lodash.sortby');
+clone = require('lodash.clone');
 
 propertyCheckList = yaml.safeLoad(fs.readFileSync(path.join(__dirname, '../../data', 'properties.yml'), 'utf8')).split(' ');
 
@@ -71,6 +78,95 @@ sortProperties = function (properties, order) {
   }
 };
 
+trimGroupSeparator = function (subject) {
+  if (!subject.length) {
+    return subject;
+  }
+
+  if (subject[0] === GROUP_SEPARATOR) {
+    subject.shift();
+
+    return trimGroupSeparator(subject);
+  }
+
+  if (subject[subject.length - 1] === GROUP_SEPARATOR) {
+    subject.pop();
+
+    return trimGroupSeparator(subject);
+  }
+
+  return subject;
+};
+
+validateSeparation = function (parser, block, expectedPropertyOrder) {
+  var blockProperties;
+
+  blockProperties = [];
+
+  block.forEach(function (dec) {
+    var prop,
+        name;
+
+    if (dec.type === 'declaration') {
+      prop = dec.first('property');
+      name = prop.first('ident');
+
+      if (!name) {
+        return;
+      }
+
+      if (parser.options['ignore-custom-properties'] && propertyCheckList.indexOf(name.content) === -1) {
+        return;
+      }
+
+      blockProperties.push(name.content);
+    }
+
+    if (dec.type === 'space' && dec.content.indexOf('\n\n') === 0) {
+      blockProperties.push(GROUP_SEPARATOR);
+    }
+  });
+
+  expectedPropertyOrder = clone(expectedPropertyOrder);
+
+  // First remove expectedPropertyOrder properties that do not appear in blockProperties.
+  expectedPropertyOrder = expectedPropertyOrder.filter((propertyName) => {
+    if (propertyName === GROUP_SEPARATOR) {
+      return true;
+    }
+
+    return blockProperties.indexOf(propertyName) !== -1;
+  });
+
+  expectedPropertyOrder = trimGroupSeparator(expectedPropertyOrder);
+
+  // Then do the inverse with blockProperties.
+  blockProperties = blockProperties.filter((propertyName) => {
+    if (propertyName === GROUP_SEPARATOR) {
+      return true;
+    }
+
+    return expectedPropertyOrder.indexOf(propertyName) !== -1;
+  });
+
+  blockProperties = trimGroupSeparator(blockProperties);
+
+  blockProperties = blockProperties.join(', ');
+  expectedPropertyOrder = expectedPropertyOrder.join(', ');
+
+  if (blockProperties !== expectedPropertyOrder) {
+    return {
+      'ruleId': parser.rule.name,
+      'line': block.start.line,
+      'column': block.start.column,
+      'message': 'Expected "' + expectedPropertyOrder + '", found "' + blockProperties + '"',
+      'severity': parser.severity
+    };
+  }
+
+  return null;
+};
+
 module.exports = {
   'name': 'property-sort-order',
   'defaults': {
@@ -79,6 +175,7 @@ module.exports = {
   },
   'detect': function (ast, parser) {
     var result = [],
+        validateSeparationResult,
         order = getOrderConfig(parser.options.order) || parser.options.order;
 
     ast.traverseByType('block', function (block) {
@@ -128,6 +225,14 @@ module.exports = {
           'message': 'Expected "' + expectedOrder + '", found "' + currentOrder + '"',
           'severity': parser.severity
         });
+      } else {
+        if (typeof order === 'object') {
+          validateSeparationResult = validateSeparation(parser, block, order);
+
+          if (validateSeparationResult) {
+            result = helpers.addUnique(result, validateSeparationResult);
+          }
+        }
       }
     });
 
--- a/lib/rules/property-sort-order.js
+++ b/lib/rules/property-sort-order.js
@@ -225,7 +225,8 @@ module.exports = {
           'message': 'Expected "' + expectedOrder + '", found "' + currentOrder + '"',
           'severity': parser.severity
         });
-      } else {
+      }
+      else {
         if (typeof order === 'object') {
           validateSeparationResult = validateSeparation(parser, block, order);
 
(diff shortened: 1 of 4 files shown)


lib/rules/property-sort-order.js
```diff
@@ -1,89 +1,74 @@
...
-
-    for (i = 0; i < order.length; i++) {
-      if (keys.indexOf(order[i]) !== -1) {
-        orderedKeys.push(order[i]);
-      }
+      return sortBy(properties, 'name');
     }
 
-    keys = orderedKeys;
+    throw new Error('Unexpected sorting order.');
```
benthemonkey: As far as I know we haven't been throwing errors on incorrect configuration options. I think for consistency this should fail silently.

gajus: If you were doing something horribly wrong, that does not mean you should continue doing it. Thats just terrible advise.

benthemonkey: I grepped our code base and the only `throw`s are in `index.js`, `lib/groot.js` and `lib/rules.js`

lib/rules/property-sort-order.js
```diff
@@ -110,35 +95,51 @@ module.exports = {
...
+        expectedOrder = sorted.map(function (prop) {
+          return prop.name;
+        }).join(', ');
+
+        currentOrder = properties.map(function (prop) {
+          return prop.name;
+        }).join(', ');
+
+        // console.log('expectedOrder', expectedOrder);
+        // console.log('currentOrder', currentOrder);
```
benthemonkey: These comments should be removed before the PR gets merged.

gajus: @benthemonkey It passes ESLint and there is no mention of this requirement in https://github.com/sasstools/sass-lint/blob/develop/CONTRIBUTING.md. It is left there for the convenience of the potential contributors.

gajus:
The latest commit enables group separation.

https://github.com/gajus/sass-lint/commit/3f6be12618217b535236bd8ad56807f642f5570d

I like to be explicit over implicit, therefore, I did not blindly follow scss-lint syntax for declaring group separation. User needs to explicitly say '(group separator)' in their configuration, e.g.

```yml
property-sort-order:
  - 1
  - order:
      - composes
      - '(group separator)'
      - font
      - font-family
      - font-size
      - font-weight
      - font-style
      - font-variant
      - font-size-adjust
      - font-stretch
      - font-effect
      - font-emphasize
      - font-emphasize-position
      - font-emphasize-style
      - font-smooth
      - line-height
      - '(group separator)'
      - position
      - z-index
      - top
      - right
      - bottom
      - left
```

It is an arbitrary choice. I am open to discuss alternative naming convention.

benthemonkey:
I like the thought of being explicit. I don't think you need the single quotes around `(group separator)` for it to get parsed (at least when I tested your example on the js-yaml demo page).

As another option, the user could pass nested arrays:
```yaml
property-sort-order:
  - 1
  - order:
    - group 1:
      - composes
    - group 2:
      - font
      - font-family
      - font-size
    - group 3:
      - position
      - z-index
      - top
      - right
      - bottom
      - left
```

The group names would be optional for clarity. I like the visual structure of the added nesting, but I don't like that it would cause there to be two formats for the `order` option, which is probably bad practice.

BPScott:
Personally, I prefer the idea of nested groups over the magic string.

> it would cause there to be two formats for the order option, which is probably bad practice.

My gutfeel is that two formats are more explicit, maintainable and easier to switch between than having a single format, but with an arbitrarily (some might say magic) string delimiting blocks to trigger the group block behaviour. Our of the two "that feels odd"s (multiple syntaxes vs magic strings) the syntax change feels like the lesser of two gotchas.

But hey I'm just some dude on the internet who isn't a maintainer, my reckons hold no intrinsic weight :)

DanPurdy:
@Snugug @bgriffith what do you guys reckon? Nested arrays seems like the most sensible option here for me, Im not sure they should be named as above but i think it makes more sense than adding a keyword that could be spelt wrong or would need be verified.

Snugug:
Agreed on nested arrays. Don't like magic spaces

gajus:
Even if choose nested arrays over a magic string:

* The actual implementation will require to flatten the array and rely on a magic string/symbol.
* Even if we do nested array, we need a "magic" string to report the error, e.g.

```
Expected "width, height, (group separator), display, (group separator), border", found "width, height, (group separator), display, border"
```

I think having the same string appear in the error and in the rule declaration makes the error itself more clear.

gajus:
Another argument pro magic string is that magic strings allow more flexible definition of the requirements, e.g. you could require that there are two spaces between certain groups.

```yml
- order:
      - composes
      - '(group separator)'
      - '(group separator)'
      - font
```

Snugug:
It's critically important that we separate implementation details from user experience. What provides the most clear and consistent UX is what we should be aiming for, and figure out how to accommodate that with our implementation. As for group separator, that's trivial to document.

Nested arrays provide clear indicators (that are impossible to miss due to the indentation) to the user, as well as provides a way for users to name their groupings, which we can utilize for better error handling, even better than separator. In addition, doing this through groups means that we have logical, named groups for when we implement https://github.com/sasstools/sass-lint/issues/350 to inform users where properties should be placed in.

Finally, enforcing groups as opposed to a magic string will specifically prevent users from doing what you've shown in https://github.com/sasstools/sass-lint/pull/423#issuecomment-158781204, putting multiple separators next to each other   which doesn't make much logical sense. Nested arrays forces the user to only create logical, named groupings through the syntax, which is a UX plus.

Please update to move from magic strings to nested groups.

gajus:
Please advise what should be the error message phrasing.

Snugug:
There are now multiple error states that we should be addressing:

* _No Groups_ - Properties should be sorted in the following order: {{properties}}
* _Groups - No Spaces_ - Groups {{name} and {{name}} should be separated by a new line
* _Groups - Property In Wrong Group_ - Property {{property}} is in the wrong group. Please move it to {{group}}
* _Groups - Properties Not Sorted_ - Properties ing {{group}} should be sorted in the following order: {{properties}}
* _Groups - Unknown Property in Group_ - Property {{property}} should be placed after all property groups
* _Groups - Unknown Property Group Without Space_ - Group {{last group in block}} should be followed by a new line before ungrouped properties are included

This provides the most accurate reporting to our users. In addition, our example in our documentation and tests should be something akin to [MDO's Declaration Order](http://mdo.github.io/code-guide/#css-declaration-order) to help inform our users that they should use logical names for their groups.

gajus:
> Groups - No Spaces - Groups {{name} and {{name}} should be separated by a new line

(and all the other error variations that contain group name)

assume that person knows by memory what are the names of groups, what is the order of groups, and what properties each group must contain. This is nonsensical – it introduces a huge learning barrier.

Other errors, e.g.

> Groups - Property In Wrong Group - Property {{property}} is in the wrong group. Please move it to {{group}}

Give only partial information about the mistake. It tells me that my property is in the wrong group, in which group it should be, but does not say where in the said group it should be.

Finally, having 6 different error messages for what is essentially covered by a single message, e.g.

> Expected "width, height, (group separator), display, (group separator), border", found "width, height, (group separator), display, border"

introduces additional mental strain when it comes to reading the lint report.

Snugug:
The misalignment we seem to be having is between bluntness and precision. Sass Lint errors on the side of precision, not bluntness.

Keep in mind that each one of these errors will include line and column number, so the question of "where is this coming from" is moot. Where is this going is important to figure out (maybe include line:column for group to move to).

As for the argument that a full list of properties is easier, that is only accurate if every single property is a block is unsorted. If a group separator is missing, or if properties in the middle or end are unsorted, it becomes very hard to scan that whole list and identify the issue. 

Considering how complex this rule is, how many different configuration options there are, and how much information the errors need to convey to our users, emitting multiple error messages to convey that complexity in precise means is very important.

gajus:
> The misalignment we seem to be having is between bluntness and precision. Sass Lint errors on the side of precision, not bluntness.

I think this works with all of the rules except `property-sort-order`, e.g.

afc24500-8eff-11e5-8b7f-e886c8cab468

This makes it look like a cyclic error. In fact, I had people complaining about it (https://github.com/gajus/canonical/issues/3).

> Keep in mind that each one of these errors will include line and column number, so the question of "where is this coming from" is moot. Where is this going is important to figure out (maybe include line:column for group to move to).

Line number only makes sense in the context of a block. In the context of a property order, specific property line number has 0 value, e.g.

```js
{
    "filePath": "property-sort-order.scss",
    "warningCount": 4,
    "errorCount": 0,
    "messages": [
        {
            "ruleId": "property-sort-order",
            "line": 2,
            "column": 3,
            "message": "Expected `height`, found `composes`",
            "severity": 1
        },
        {
            "ruleId": "property-sort-order",
            "line": 3,
            "column": 3,
            "message": "Expected `width`, found `height`",
            "severity": 1
        },
        {
            "ruleId": "property-sort-order",
            "line": 5,
            "column": 3,
            "message": "Expected `border`, found `width`",
            "severity": 1
        },
        {
            "ruleId": "property-sort-order",
            "line": 6,
            "column": 3,
            "message": "Expected `composes`, found `border`",
            "severity": 1
        }
    ]
}
```

As soon as I address the first warning, all the following error message line numbers are misaligned.

Snugug:
Please do re-read https://github.com/sasstools/sass-lint/pull/423#issuecomment-158915062. I'm not suggesting that we keep per-property sorting (FWIW, a single person getting confused is not "people are complaining about it"), but rather that wrapping up all of the complexity this rule is now encompassing into a single complex error statement is a bad idea.

Snugug:
I'm actually missing a state in that comment: _Group - Group Out Of Order_

gajus:
I am not going to proceed with this PR because I don't agree with the proposed behaviour of the rule. I will release this particular implementation as a plugin as soon as it becomes possible, https://github.com/sasstools/sass-lint/issues/139. In the mean time, I am happy for someone else to takeover this PR and carry the developed in the proposed direction.

gajus:
In the mean time, I have made a branch that includes this particular PR and released it under https://www.npmjs.com/package/css-lint name.

PR #436, opened 2015-12-02
Set severity and options at the top of function

bquorning:
Checking `typeof fullRule` 1 time up front instead of 3-4 times throughout the function seems cleaner.

DCO 1.1 Signed-off-by: Benjamin Quorning <dco@quorning.net>

### DIFF
--- a/lib/rules.js
+++ b/lib/rules.js
@@ -27,21 +27,30 @@ module.exports = function (config) {
   Object.keys(config.rules).forEach(function (rule) {
     var fullRule = config.rules[rule],
         loadRule,
+        severity,
         options,
         ruleSearch;
+
+    if (typeof fullRule === 'number') {
+      severity = fullRule;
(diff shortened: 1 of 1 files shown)


DanPurdy:
Hi @bquorning Is there an issue open for this? We usually only accept PR's based on issues so that everyone can agree on the right way forward.

I'll leave this one for @Snugug to have a look

bquorning:
No, there is no open issue. I was just reading through the source and found this part a bit hard to understand.

benthemonkey:
This code seems much more readable to me, so I'm in favor of it. :+1:

DanPurdy:
I'm ok with this once it's updated to the latest develop. Thanks!

bquorning:
Rebased on origin/develop.

DanPurdy:
Thanks @bquorning

PR #440, opened 2015-12-07
Adjust no-trailing-zero to fail on first decimal

### LINKED ISSUE #438: New Rule: Unnecessary Mantissa
Name: `unnecessary-mantissa`

Add a rule that enforces that removal of all unnecessary trailing decimal points (e.g. `4.0` should be written as `4`).

I was a little ambitious and have already written the code to complete this as [`a7db54b`](https://github.com/Dru89/sass-lint/commit/a7db54b6b910e3f7ce6623a2c1d9eca7c5fcc017) but will wait to submit the pull request as per the contributing guidelines.

### LINKED ISSUE #439: Trailing zero issue
As discovered in #438 it seems that the `no-trailing-zero` rule only works correctly when there is at least one non zero number following a decimal point.

The following is currently valid.
```scss

foo {
  margin: 1.0000000000
}
```

The following is not

```scss

foo {
  margin: 1.10000000000
}
```

Dru89:
The `no-trailing-zero` rule was not failing on numbers like `1.0`, as the regular expression required there to be a non-zero number in the decimal for the check to work.  

This change removes that constraint and adds test cases to make sure that the result works as expected.  The change fixes #438 and fixes #439.

DCO 1.1 Signed-off-by: Drew Hays <drewshays@gmail.com>

### DIFF
--- a/lib/rules/no-trailing-zero.js
+++ b/lib/rules/no-trailing-zero.js
@@ -2,7 +2,7 @@
 
 var helpers = require('../helpers');
 
-var trailingZeroRegex = /^(\d+\.|\.)+(\d*?[1-9])0+$/;
+var trailingZeroRegex = /^(\d+\.|\.)+(\d*?)0+$/;
 
 module.exports = {
   'name': 'no-trailing-zero',
--- a/docs/rules/no-trailing-zero.md
+++ b/docs/rules/no-trailing-zero.md
@@ -18,4 +18,8 @@ When enabled, the following are disallowed:
 .foo {
   margin: 0.2500rem;
 }
+
+.foo {
+  margin: 4.0rem;
+}
 ```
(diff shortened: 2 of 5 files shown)


Dru89:
Welp, that was embarrassing.  I kept playing with the test cases to make sure that I felt everything was sufficiently covered, but forgot to commit the changes to the test file.  It should be updated now.

benthemonkey:
LGTM :+1:

DanPurdy:
Thanks for this!

DanPurdy:
Yep, just checked it against some other values, all good to me, thanks @Dru89

lib/rules/no-trailing-zero.js
```diff
@@ -2,7 +2,7 @@
 
 var helpers = require('../helpers');
 
-var trailingZeroRegex = /^(\d+\.|\.)+(\d*?[1-9])0+$/;
+var trailingZeroRegex = /^(\d+\.|\.)+(\d*?)0+$/;
```
bquorning: I believe `^(\d+\.|\.)` could be changed to `^(\d*\.)` which in my opinion is a bit easier to read.

Issue #435, opened 2015-11-30
Negative percentage values still flagged by shorthand-values


attila:
This could be related to https://github.com/sasstools/sass-lint/pull/429 but

```scss
.foo {
  margin: -10% -20%;
}
```

is flagged by "shorthand-values" in latest develop.

Command output:

```
$ ./$ node_modules/.bin/sass-lint -v test.scss

test.scss
  2:11  warning  Property `margin` should be written more concisely as `-` instead of `- -`  shorthand-values

✖ 1 problem (0 errors, 1 warning)
```

DanPurdy:
Confirmed, looks like something to do with percentage values and their structure in gonzales. Will look into it, thanks @attila .

DanPurdy:
sorted in #441 will be part of the 1.4.0 release scheduled for this week

PR #441, opened 2015-12-07
Fix issue with shorthand percentage values

DanPurdy:
Fixes an issue with the `shorthand-values` rule as highlighted in #435

Also adds a sample of the yml configuration to the docs as I felt the syntax was useful to see examples of for those not too familiar with YAML.

I chose not to bundle this as a hotfix as I don't think there's any point trying to push 1.3.4 out before 1.4.0 is released..

` DCO 1.1 Signed-off-by: Dan Purdy danjpurdy@gmail.com `


DanPurdy:
closes #435

bgriffith:
Looks good!

PR #446, opened 2015-12-10
Bem Depth remove stray console.logs

DanPurdy:
Dont know how but some console logs managed to slip through with the recent bem-depth rule, my bad!

Anyway this just removes them...


benthemonkey:
Oops!

PR #451, opened 2015-12-13
Fix 'ignore' setting

michalsnik:
When trying to set 'ignore' some files from linting, current version of sass-lint doesn't get it, and lint them anyway. I'm checking if current file in 'lintText()' is in configs ignore list, and prevent from linting it.

DCO 1.1 Signed-off-by: Michał Sajnóg <michal.sajnog@hotmail.com>

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,8 +37,8 @@
     "lodash.capitalize": "^3.0.0",
     "lodash.kebabcase": "^3.0.1",
     "merge": "^1.2.0",
-    "util": "^0.10.3",
-    "minimatch": "^3.0.0"
+    "minimatch": "^3.0.0",
+    "util": "^0.10.3"
   },
   "devDependencies": {
     "coveralls": "^2.11.4",
--- a/package.json
+++ b/package.json
@@ -37,7 +37,8 @@
     "lodash.capitalize": "^3.0.0",
     "lodash.kebabcase": "^3.0.1",
     "merge": "^1.2.0",
-    "util": "^0.10.3"
+    "util": "^0.10.3",
+    "minimatch": "^3.0.0"
   },
   "devDependencies": {
     "coveralls": "^2.11.4",
--- a/index.js
+++ b/index.js
@@ -32,14 +32,16 @@ sassLint.resultCount = function (results) {
 };
 
 sassLint.lintText = function (file, options, configPath) {
-  var rules = slRules(this.getConfig(options, configPath)),
+  var config = this.getConfig(options, configPath),
+      rules = slRules(config),
       ast = groot(file.text, file.format, file.filename),
       detects,
       results = [],
       errors = 0,
-      warnings = 0;
+      warnings = 0,
+      ignores = config.files.ignore || '';
 
-  if (ast.content.length > 0) {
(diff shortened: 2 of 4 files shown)


DanPurdy:
Hi @michalsnik can you please read and follow the contribution guidelines before submitting PR's 

https://github.com/sasstools/sass-lint/blob/develop/CONTRIBUTING.md#pull-requests

Thanks

michalsnik:
Hi,
Sure, i'll create appropriate issue later.

michalsnik:
I updated my PR, now it solves issue #452 

I'm checking if currently linted file matches any of 'ignore' setting value, either string and array. To check if it matches I added minimatch package, which is also used by glob package.
What do you think about it? Should I write some tests for this? Or I simply don't understand something in sass-lint and everything works fine without my change?

My PR makes it possible to set for example:
zrzut ekranu 2015-12-13 23 34 50
zrzut ekranu 2015-12-14 00 15 41
zrzut ekranu 2015-12-14 00 16 00

michalsnik:
I moved my code into new helper function and wrote some tests.

package.json
```diff
@@ -37,7 +37,8 @@
     "lodash.capitalize": "^3.0.0",
     "lodash.kebabcase": "^3.0.1",
     "merge": "^1.2.0",
-    "util": "^0.10.3"
+    "util": "^0.10.3",
+    "minimatch": "^3.0.0"
```
benthemonkey: Can you please move this to keep our dependencies in alphabetical order?

benthemonkey:
I had a tiny comment then I think this is ready to merge. @DanPurdy any reason to wait?

DanPurdy:
Sorry, I've just not had the time to look into anything this week, will go over this today.

michalsnik:
@benthemonkey thanks, i just reordered dependencies

DanPurdy:
OK @benthemonkey & @michalsnik I don't think we can merge this, I need to have a word with @Snugug because for me the issue isn't actually with `sass-lint` it's with `gulp-sass-lint`. `gulp-sass-lint` is using the `lintText()` method when it probably should be using the `lintFiles()` method.

We already load the config and sort out all of the ignores and includes including handling all glob patterns in `lintFiles` which then fires into `lintText` so we'd end up duplicating a lot of functionality for people using the CLI, `grunt-sass-lint` and some of the IDE integrations that are about.

Snugug:
@DanPurdy and I talked about this; it does need to be resolved in `lintText` not `gulp-sass-lint`, but we need to look in to this more to see if this is the way we'd like it resolved.

DanPurdy:
For reference in the future, I handled this in `linter-sass-lint` for atom with the globule module 
https://github.com/AtomLinter/linter-sass-lint/blob/master/lib/main.coffee#L133

There could be a similar approach taken in gulp/grunt sass-lint as long as we can relativise the paths between the config and what gulp is receiving or we could load the config once in lintText, cache-it and then check for file matches before actually applying our rules.

krzysztofnowak:
Hi Guys,
Any progress with this issue?

Issue #221, opened 2015-09-25
CLI should provide a list of errors and exit with the appropriate exit code


BPScott:
Using Sass-lint 1.2.2.

I'd like to use the CLI for my development workflow, running sass-lint to check quality. I'd like to:
1) Get a list of issues in the CLI's output, so I can see an overview of all issues rather than having to fix them one at a time.
2) Have `sass-lint` to exit with code 0 on success, and 1 if there are any errors.

Currently I can either:

* Run `sass-lint -v` which throws an exception on the first error but does exit with code 1. This fails the first case but passes the second.
* Run `sass-lint -vq` which provides a list of all errors, but always exits with code 0. This passes the first case but fails the second

Neither of which let me do what I'm hoping. I'm not sure what the best solution for this would be. Taking a quick look at eslint, when running it without options, they output the full list of issues every time, if sass-lint were to mimic this then the default action would be akin to currently running `sass-lint -qv`, but also providing the proper exit code.

DanPurdy:
Hey @BPScott so I've definitely got this in mind right now..  Once I've finished writing the rules that I have left on my plate this will be my next focus. 

I completely agree that proper error codes need to be implemented soon and functionality should mimic to a point other successful linters, I think our main focus at the moment is getting our rules and our original goal of supporting both the `scss` and `Sass` syntax sorted and then we're looking into improving functionality. We are well on our way though!

BPScott:
Hi @DanPurdy, indeed it's amazing to see sass-lint hurtling along at such a great pace. Figured I'd raise an issue just to make sure this doesn't get lost down the sofa tracked. Totally agree that beefing out the ruleset is top priority :)

DanPurdy:
Ha thanks @BPScott we'll try and keep the pace up! :smile:

bquorning:
I’m using this shell script as a work-around until sass-lint have proper exit status codes:
```sh
SASS_LINT_OUTPUT=$(node_modules/.bin/sass-lint -v)

if [ -n "$SASS_LINT_OUTPUT" ]; then
  echo "$SASS_LINT_OUTPUT"
  exit 1
fi
```

kmiyashiro:
This should be closed, it was fixed in #458

glen-84:
@kmiyashiro,

> This doesn't fully solve the issues raised there as I believe we should be supporting a range of different error codes but this for now solves the main problem that's been reported.

mvz:
[...]

PR #458, opened 2015-12-20
Update to error codes

DanPurdy:
As discussed in #221 when errors are detected from the CLI there is an inconsistent error code being thrown.

This doesn't fully solve the issues raised there as I believe we should be supporting a range of different error codes but this for now solves the main problem that's been reported.

Took the opportunity to improve comments and structure within the main index file too.

If any errors are detected now the CLI will always exit with a error code of 1 even if errors are supressed with the `--quiet` flag, warnings will still return a code of 0.

@BPScott you may want to double check this meets your requirements?

### DIFF
--- a/bin/sass-lint.js
+++ b/bin/sass-lint.js
@@ -7,7 +7,8 @@ var program = require('commander'),
 
 var configPath,
     ignores,
-    configOptions = {};
+    configOptions = {},
+    exitCode = 0;
 
 var detectPattern = function (pattern) {
   var detects;
@@ -16,6 +17,10 @@ var detectPattern = function (pattern) {
 
   if (program.verbose) {
     lint.outputResults(detects, configOptions, configPath);
+  }
+
+  if (lint.errorCount(detects).count) {
+    exitCode = 1;
   }
 
   if (program.exit) {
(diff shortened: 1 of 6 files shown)


DanPurdy:
node v0.10.x has an issue with child process it seems where if an unhandled error is detected (i.e. the ones we throw with error counts in them if not set to `--no-exit`) it throws an error code of 8, I've verified that error code 1 is still output when not in a child process, unless we want a test case that check for ode versions too ive had to remove the error code test for non quiet output.

DanPurdy:
@bgriffith I went back and double checked this. The idea was to make sure that an exit code of 1 was thrown when the `--no-exit` flag was set on the CLI.

Currently even if errors are detected an exit code of 0 is thrown when set to `--no-exit` this is causing problems for those who use this as a CI tool as it will report their builds as passing when in fact errors are present.

Warnings still display with a code of 0

You can test this out yourself if you would like, go into the sass-lint directory and run 
`sass-lint --config tests/yml/.error-output.yml tests/sass/id-name-format.scss --verbose --no-exit`
or 
`sass-lint --config tests/yml/.error-output.yml tests/sass/id-name-format.scss --verbose -q`

then once you see the output in bash, type `echo $?` which will show you the last exit code.

:+1: 

I would say it's good to merge if this passes code review.

bgriffith:
Looks good! Tests worked!

PR #459, opened 2015-12-21
Return parse errors as lint errors

DanPurdy:
In an effort to reduce the amount of unhandled errors thrown by `sass-lint` I've converted the parse errors and undefined tree errors that can be thrown while the AST is being constructed into lint errors.

Firstly this prevents AST bugs from crashing peoples linting builds altogether and only prevents the file with the parse error from being linted rather than the whole project.

Secondly as the line is reported directly as a parse error it should allow people to provide a more useful bug report in the event of something un expected happening rather than them having to be confronted with a meaningless stack trace at the moment. 

I've gone with the Eslint report style in the sense that the ruleId reported for these errors is `Fatal`.

There's not a specific issue that this relates to for `sass-lint` other than my desire to remove all thrown errors as best as possible but it does relate to many issues encountered in `gulp-sass-lint` and `grunt-sass-lint` etc

[#18](https://github.com/sasstools/gulp-sass-lint/issues/18) for example.

Errors thrown by failing rules will still be thrown as errors for now as they do break the lint cycle.

Not the cleanest implementation right now but I figure this will be a big consideration once we move to v2.0.

Also added a few tests to the CLI to ensure these errors are reported correctly.

### DIFF
--- a/index.js
+++ b/index.js
@@ -33,13 +33,29 @@ sassLint.resultCount = function (results) {
 
 sassLint.lintText = function (file, options, configPath) {
   var rules = slRules(this.getConfig(options, configPath)),
-      ast = groot(file.text, file.format, file.filename),
+      ast = {},
       detects,
       results = [],
       errors = 0,
       warnings = 0;
 
-  if (ast.content.length > 0) {
+  try {
+    ast = groot(file.text, file.format, file.filename);
+  }
+  catch (e) {
+    errors++;
+    var line = e.line || 1;
+
+    results = [{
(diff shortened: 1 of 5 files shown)


DanPurdy:
Damn you coveralls! :stuck_out_tongue: 

I'll add some non cli unit tests too later on to check the results.

DanPurdy:
Oops, just had a moment there..

bgriffith:
Looks good - will solve one of the major gripes at the mo.

PR #504, opened 2016-02-05
Gonzales 3.2 - Fix spaceBeforeBrace

danez:
Fixes the `space-before-brace` rule to work with the latest version of gonzales.

node.last() changed from returning undefined to null. As ```typeof null === object``` the check was failing.
The new check is now accounting for all falsy values, and as false is captured in a separate check the logic is still the same.

### DIFF
--- a/lib/rules/space-before-brace.js
+++ b/lib/rules/space-before-brace.js
@@ -7,7 +7,7 @@ var getLastWhitespace = function (node) {
     return null;
   }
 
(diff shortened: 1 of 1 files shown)


bgriffith:
Awesome thanks! Could you just add the `Developer Certificate of Origin` thanks

https://github.com/sasstools/sass-lint/blob/develop/CONTRIBUTING.md#pull-requests

PR #512, opened 2016-02-07
Gonzales 3.2 - Fix gonzales-pe version to 3.2.2

danez:
see tonyganch/gonzales-pe#149

Currently I'm not able to install versions > 3.2.2

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "eslint": "^1.1.0",
     "fs-extra": "^0.26.0",
     "glob": "^6.0.0",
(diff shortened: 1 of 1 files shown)


bgriffith:
No longer needed as 3.2.6 now installs ok :+1:

DanPurdy:
I think we should definitely fix our versions once we release though. We can setup greenkeeper.io to do the updates for us rather than see different versions with possible regressions break.

Snugug:
I've just enabled Greenkeeper

DanPurdy:
:+1:

PR #513, opened 2016-02-07
Gonzales 3.2 - Fix rule nesting-depth

danez:
This fixes the rule `nesting-depth`.

Node type `simpleSelector` has been removed.

Currently only the scss part is fixed, sass throws Fatal.


danez:
As of gonzales 3.2.6 also the sass tests now work.

PR #515, opened 2016-02-07
Add node version

DanPurdy:
Checking the node versions, we aren't testing against node v4 anymore..

https://github.com/nodejs/LTS#example has the dates of when LTS support will end. So in October we should remove support/testing for node 0.10.x and in December remove support for 0.12.x


Snugug:
LGTM :+1:

PR #526, opened 2016-02-13
Gonzales 3.2 - Fix attribute nesting

DanPurdy:
Fixes the `force-attribute-nesting` - rule to work with the latest version of gonzales.

Note: Travis will fail as there are broken tests on gonzales-3-develop branch. Verify by checking force-attribute-nesting rule test success.


benthemonkey:
:+1: I love how much shorter this is now.

DanPurdy:
Definitely, the whole way that gonzales-pe 3.2 handles selectors is much much clearer imo. Don't have to worry about checking for delimiters and such like anymore.

PR #541, opened 2016-02-20
Fix dot in filename issue with clean-import-paths rule

### LINKED ISSUE #536: False positive for clean-import-paths with dot in the filename
When I try to import a file that has a dot in it's filename, the rule clean-import-paths rule throws a warning.

```scss
@import '../../node_modules/inuit-normalize/generic.normalize';
```

The name of the file is `_generic.normalize.scss` — so the clean import path above is correct, or am I missing something? Maybe the rule should just check for scss/sass file extensions?

bgriffith:
Fix an issue with dots in filenames.

Closes #536
 

DanPurdy:
:+1:

PR #582, opened 2016-03-28
Gonzales 3 update eslint 2.x

DanPurdy:
Seen as we are waiting for gonzales to allow us to release 1.6 I thought I'd jump ahead and just get these eslint updates done too.

Updates our config with the latest format of the rules for eslint as some of the rules have merged/changed names in version 2.x, It also fixes all of the issues that eslint now flags up on our shiny 1.6 branch so it should all be ready to be released once we go ahead with it.

`<DCO 1.1 Signed-off-by: Dan Purdy danjpurdy@gmail.com>`


DanPurdy:
Oops wrong branch

PR #584, opened 2016-03-28
Fix valid default properties - Gonzales 3.2

### LINKED ISSUE #571: variable-for-property rule should allow reserved words
**Issue**
The rule `variable-for-property` reports reserved words as being non-variables (which is technically true of course). This means it's impossible to use words like `transparent` or `none`.

**Possible solution**
Make a list of reserved words that are allowed when using `variable-for-property` (equal to [scss-lint](https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#variableforproperty), see [this line](https://github.com/brigade/scss-lint/blob/e9a8e243b43bc4fe30a919ebea2e413087546a95/lib/scss_lint/linter/variable_for_property.rb#L6)).

**Example**
This can be tested with the following config:

```yml
options:
  merge-default-rules: false
  formatter: stylish
rules:
  variable-for-property:
    - 2
    -
      properties:
        - background
```

And this CSS:

```css
.test {
  background: inherit;
  background: initial;
  background: transparent;
  background: none;
}
```

Which generates this errors when using sass-lint v1.5.1:

```
test.scss
  2:3  error  Values for properties of type 'background' may only be variables  variable-for-property
  3:3  error  Values for properties of type 'background' may only be variables  variable-for-property
  4:3  error  Values for properties of type 'background' may only be variables  variable-for-property
  5:3  error  Values for properties of type 'background' may only be variables  variable-for-property

✖ 4 problems (4 errors, 0 warnings)
```

DanPurdy:
Fixes #571 

The variable for property rule wasn't taking into account reserved words usage.

This fix implements a function to check the property in question against a list of whitelisted ident values as well as it's type as is currently happening.

`<DCO 1.1 Signed-off-by: Dan Purdy danjpurdy@gmail.com>`

### DIFF
--- a/docs/rules/variable-for-property.md
+++ b/docs/rules/variable-for-property.md
@@ -1,6 +1,12 @@
 # Variable For Property
 
-Rule `variable-for-property` will enforce the use of variables for the values of specified properties. There are no properties by default.
+Rule `variable-for-property` will enforce the use of variables for the values of specified properties.
+There are no properties by default, except for reserved words listed below which are always whitelisted:
+* inherit
+* initial
+* transparent
+* none
+* currentColor
 
 ## Options
 
--- a/lib/rules/variable-for-property.js
+++ b/lib/rules/variable-for-property.js
@@ -2,6 +2,27 @@
 
 var helpers = require('../helpers');
 
+// The whitelisted ident values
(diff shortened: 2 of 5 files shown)


bgriffith:
:+1:

PR #590, opened 2016-03-29
Temporary gonzales

DanPurdy:
Due to the issue here https://github.com/tonyganch/gonzales-pe/issues/149 gonzales-pe can no longer be installed on windows. As a temporary stop gap I've removed the post install script and just updated a few things and republished to npm as gonzales-pe-sl.

This PR replaces it in sass-lint and runs all our tests fine. If we could test this on a windows machine now that would be great.. I dont have access or time to one tonight!

### DIFF
--- a/lib/groot.js
+++ b/lib/groot.js
@@ -3,7 +3,7 @@
 //////////////////////////////
 'use strict';
 
-var gonzales = require('gonzales-pe');
+var gonzales = require('gonzales-pe-sl');
 
 module.exports = function (text, syntax, filename) {
   var tree;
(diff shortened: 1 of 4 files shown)


DanPurdy:
We can switch back to the official gonzales-pe once it's updated to remove the bash-centric post install script

mven:
Installed just fine for me on:

Windows 7
NPM v2.14.20
Node v4.4.0

screen shot 2016-03-29 at 7 32 29 pm

DanPurdy:
Thanks @mven!

DanPurdy:
Verified to be installable on windows and non bash systems.. Made me aware of some Sass format windows issues though, mostly look to be AST related again though..

PR #605, opened 2016-04-12
New rule: no-trailing-whitespace

### LINKED ISSUE #596: Trailing whitespace rule
I couldn't find a rule that prohibits trailing whitespace at the end of a line:

    .foo {\s\s\s
        color: red;\s\s
    }

Is there such a rule, and if not, does it make sense in the context of sass-lint?

gyoshev:
Will fix #596 

Do I need to add new rules to `lib/config/sass-lint.yml`?

DCO 1.1 Signed-off-by: Alex Gyoshev alex@gyoshev.net

### DIFF
--- a/lib/config/sass-lint.yml
+++ b/lib/config/sass-lint.yml
@@ -30,6 +30,7 @@ rules:
   no-mergeable-selectors: 1
   no-misspelled-properties: 1
   no-qualifying-elements: 1
+  no-trailing-whitespace: 1
   no-trailing-zero: 1
   no-transition-all: 1
   no-url-protocols: 1
--- a/docs/rules/no-trailing-whitespace.md
+++ b/docs/rules/no-trailing-whitespace.md
@@ -1,0 +1,21 @@
+# No Trailing Whitespace
+
+Rule `no-trailing-whitespace` will enforce that trailing whitespace is not allowed.
+
+## Examples
+
+When enabled, the following are disallowed (\s denotes spaces or tabs):
+
+```scss
+.foo {\s
+  margin: 1.5rem;
+}
+
+.foo {
+  margin: .5rem;\s
+}
+
+.foo {
+  margin: .4rem;
+}\s
+```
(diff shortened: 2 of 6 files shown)


DanPurdy:
@gyoshev yeah, if you could add the rule to the default config that would be great, I can review the rest of this later on :+1:

gyoshev:
@DanPurdy I've added the rule to the default config, let me know if other changes are needed

lib/config/sass-lint.yml
```diff
@@ -35,6 +35,7 @@ rules:
   no-url-protocols: 1
   no-vendor-prefixes: 1
   no-warn: 1
+  no-trailing-whitespace: 1
```
DanPurdy: Just a small one but you could you put this in alphabetical order with the rest of the disallow rules so as to follow the pattern. Should be above `no-trailing-zero'.

DanPurdy:
@gyoshev  Ok this looks good, I've got one comment above but could you also add a comment above each example in the test files so that we know what part the spaces should be present for just for future reference when someone wants to update the tests as a few IDE's can be set to automatically remove trailing whitespace.

For example 
```
// Two trailing spaces after .foo
.foo {   
  margin: 2em;
```

Hopefully that makes sense, if you get those in, update with teh latest from develop and then ping me in a comment I'll merge this in for you. Thanks again! 👍

BPScott:
Probably worth fiddling with the [.editorconfig](https://github.com/sasstools/sass-lint/blob/develop/.editorconfig) so people using that don't have it automatically trim the whitespace.

DanPurdy:
Yeah I thought that but for the sake of two test files which are unlikely to be touched often it seemed a bit much, same issue with the line endings settings too. We're going to have to start adding extra tests for all the CRLF line endings due to Gonzales known issues with those.. Definitely worth looking at tweaking these configs in the future though.

The comments in the file for now will just make it obvious to everyone what is being tested within the file.

gyoshev:
Comments have been added, rule order is sorted, and the branch is rebased onto `develop`.

DanPurdy:
@gyoshev awesome! Thanks for this. I'll make an issue to investigate overwriting the editorconfig settings for these files.

👍

PR #616, opened 2016-04-19
Add Appveyor config

Snugug:
Adds Appveyor config so we can test in Windows.

No specific related issue, but we need to have CI for windows

`DCO 1.1 Signed-off-by: Sam Richard <sam@snug.ug>`

PR #618, opened 2016-04-19
Add Appveyor config

Snugug:
Adds Appveyor config so we can test in Windows.

No specific related issue, but we need to have CI for windows

`DCO 1.1 Signed-off-by: Sam Richard <sam@snug.ug>`

I totally messed up #616, let's try it again

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,29 @@
+# Test against this version of Node.js
+environment:
+  matrix:
+    - nodejs_version: '0.10'
+    - nodejs_version: '0.12'
+    - nodejs_version: '4'
+    # Latest version of Node
+    - nodejs_version: '5'
+    - nodejs_version: '1'
+
+# Install scripts. (runs after repo cloning)
+install:
+  # Get the latest stable version of Node.js or io.js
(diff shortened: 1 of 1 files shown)


Snugug:
HOW ARE YOU ALREADY OUT OF DATE?!

Either way, this should be GTG now, feel free to merge. All the Node versions are testing as expected

DanPurdy:
😄  Been trying to get all the idle PR's updated or merged so I'd just merged in a new rule that has been added.. I'm confused by these appveyor results, as i mentioned on Slack, both @bgriffith and I are seeing 40+ windows issues when we run it locally on windows but only 2 or 3 in appveyors reporting?

DanPurdy:
Would it be worth specifying an OS?

in linter-sass-lint for example i use 
`os: Windows Server 2012 R2`
can't find any info on appveyor about what their default OS is

DanPurdy:
@Snugug do you wanna update this then, I can't update it to merge it 😭

DanPurdy:
Ignoring Appveyor for now as we can move to fix the issues separately

PR #621, opened 2016-04-20
Fix space around operator

### LINKED ISSUE #572: space-around-operator should allow negative numbers without units
**Issue**
When making calculations with negative values (e.g. `width: $var / -2`) the `-` sign is incorrectly marked as an operator.

**Possible solution**
A check should be added to validate if the `-` character is used within a negative, unitless number.

**Example**
This can be tested with this config:

```yml
options:
  merge-default-rules: false
  formatter: stylish
rules:
  space-around-operator: 2
```

And this CSS:

```scss
$var: 2em;

.test {
  top: $var / -2;
  top: $var * -2;
}
```

Which produces the following errors with sass-lint v1.5.1:

```
test.scss
  4:17  error  Space expected around operator  space-around-operator
  5:17  error  Space expected around operator  space-around-operator
```

**Proposed technical solution**
In [this rule](https://github.com/sasstools/sass-lint/blob/develop/lib/rules/space-around-operator.js#L52), an exception could be added to check if `next.is('number')`. It would need to be tested to see if this satisfies all cases.

bgriffith:
Fixes an issue with the `space-around-operator` rule falsely flagging the minus operator when dealing with negative numbers.

Also fixes falsely flagging `/` in import paths for sass syntax.

Refactored rule to increase readability.

Fixes #572 


DanPurdy:
👍

PR #622, opened 2016-04-20
Fix issues with rules and cli tests

### LINKED ISSUE #445: Final newline sass issue
While looking into #437 noticed there's an issue with the final-newline rule with the sass-syntax. 

if you look [here](https://github.com/sasstools/sass-lint/blob/develop/lib/rules/final-newline.js#L19) the rule is checking for a block within the last node of the file, this isn't always possible and the error can be recreated by running the final-newline rule against our [clean-import-paths test file](https://github.com/sasstools/sass-lint/blob/develop/tests/sass/clean-import-paths.sass)

sass-lint: 1.3.3

DanPurdy:
Mainly to fix the appveyor failures, should prevent a few unhandled error throws too.

also fixes #445
`DCO 1.1 Signed-off-by: Dan Purdy <danjpurdy@gmail.com>`


DanPurdy:
So, appveyor seems a little underpowered which means i've increased the mocha timeout limit to 10 seconds to compensate for any of the file/directory reading we have to do and child process spawning as these were causing the build to fail. Also pulled a few little issues out of some of our rules and updated our test matrices. All should be passing now but we'll wait for that lovely green tick on appveyor.

This by no stretch of the imagination means sass-lint is rock solid on windows as we don't have any tests for CRLF which we know causes gonzales to behave oddly or just refuse to behave at all. Progress though at least!

bgriffith:
Looks good. Let's hope it speeds it up a little.

PR #631, opened 2016-04-21
Remove workarounds

bgriffith:
Removes workarounds that are now necessary due to fixes to gonzales.

Affects the following rules:
- `extends-before-declaration`
- `extends-before-mixins`


DanPurdy:
👍

PR #635, opened 2016-04-22
Brace style now only lints valid atrule types

### LINKED ISSUE #632: Brace style rule 'stroustrup' and multiline mixin parameters
After upgrade to 1.6.0 (to be honest I don't want to upgrade, but gulp wrapper use version ^1.5.1 - so I'm doomed...) I got many new errors in brace-styles rule.

I'm using 'stroustrup' and now this type of code fails tests:
```scss
 @mixin _suuuper-long-mixin-name(
    $_oversized-paramams-names-was-moved-to-new-line,
    $_because-i-want-to-keep-lines-under-80-chars
) {
    [...]
}
```
IMO it still fit in that rule, but tests checks just if start and end is on same line 😞

### LINKED ISSUE #634: Brace Style Throws Invalid Error
Config:
```
rules:
  brace-style:
    - 2
    -
      style: 1tbs
      allow-single-line: false
```
Example Sass (_theme.scss):
```
@import 'abstract/variables.scss';
```
Output:
```
css/sass/_theme.scss
  1:33  error  Single line statements are not allowed  brace-style
```

bgriffith:
After 1.6 `brace-style` was linting all `atrules` to check for single-line statements. Now it only lints `functions`,`if` and `else`.

Closes #634
Closes #632

### DIFF
--- a/lib/rules/brace-style.js
+++ b/lib/rules/brace-style.js
@@ -30,8 +30,20 @@ var getPreviousNode = function (node) {
     return previousParent && previousParent.contains('atkeyword') ? previousParent.last('atkeyword') : false;
   }
 
-  // Functions, Mixins, Loops
-  if (node.is('atrule') || node.is('mixin') || node.is('loop')) {
+  // Loops
+  if (node.is('loop')) {
+    return node.contains('atkeyword') ? node.last('atkeyword') : false;
+  }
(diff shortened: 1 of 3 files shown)


DanPurdy:
Looks good to me 👍

PR #646, opened 2016-04-24
Fix sass-lint 1.6 final-newline rule issues

### LINKED ISSUE #627: Sass - Nested items at the end of the file doesn't register final new line
The following code will ignore the final new line and incorrectly flag up that a new line is required. This appears only to happen when the file ends with a nested node.

```sass
.foo
  content: 'foo'
  .bar
    content: 'bar'

```
Sass 1.5.1, develop branch

### LINKED ISSUE #630: Sass - Final new line flags if anything but block at end of fine
Currently if you require a final new line, sass-lint will flag it as missing if anything bar a block is at the end.

The following causes an error. All examples have a new line at the bottom of the file.

```sass
@import test

```
```sass
$foo: true

```
```sass
$bar: (
  baz: true
)

```

Sass 1.6, develop

bgriffith:
Fixes two issues with the final-newline rule and the sass syntax, that popped up after the update to 1.6.

Fixes #627 
Fixes #630 

### DIFF
--- a/lib/rules/final-newline.js
+++ b/lib/rules/final-newline.js
@@ -2,6 +2,18 @@
 
 var helpers = require('../helpers');
 
+/**
+ * Get the 'last' node of the tree to test for an EOL
+ *
+ * @param {Object} node - The node who's last child we want to return
+ * @returns {Object} The last node
+ */
+var getLastNode = function (node) {
+  var last = node.last();
+
+  return last ? getLastNode(last) : node;
+};
+
 module.exports = {
   'name': 'final-newline',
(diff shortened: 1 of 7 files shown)


lib/rules/final-newline.js
```diff
@@ -2,29 +2,39 @@
 
 var helpers = require('../helpers');
 
+/**
+ * Get the 'last' node of the tree to test for an EOL
+ *
+ * @param {Object} node - The node who's last child we want to return
```
DanPurdy: whose

PR #539, opened 2016-02-17
Fix no-color-literals raising warning on variables named as a color

brettlangdon:
Addresses a fix for the issue outlined in #538.

DCO 1.1 Signed-off-by: Brett Langdon <me@brett.is>

PR #647, opened 2016-04-25
Update color literals

### LINKED ISSUE #538: No color literals raises warning on variable in map
I am seeing an issue where a warning is raised when a color variable is used as a map value. Specifically when the variable shares the same name as a valid CSS color.

To reproduce:

```scss
$black: #000;
$colors: (
  font-color: $black,
);
```

```bash
$ sass-lint --verbose ./test.scss
./test.scss
  3:16  warning  Color literals such as 'black' should not be used as map identifiers  no-color-literals

✖ 1 problem (0 errors, 1 warning)
```

I am running `sass-lint` `1.5.0`.

DanPurdy:
Carrying on from the changes proposed in #539 I've gone back and had a quick review of the rule.

Added two extra options
* `allow-variable-identifiers` (default `true`)
* `allow-map-identifiers` (default `true`)

The first option allows you to disallow the use of color-literals as variable identifiers/names but is disabled by default due to the feedback we received around this rule.

The second option allows you to disallow the use of color literals as map identifiers which to me seemed the logical extension of the option above but suited different use cases to the above.

Added extra tests and necessary documentation

closes #538 
closes #539

`DCO 1.1 Signed-off-by: Dan Purdy <dan@danpurdy.co.uk>`

### DIFF
--- a/lib/rules/no-color-literals.js
+++ b/lib/rules/no-color-literals.js
@@ -68,9 +68,9 @@ var checkHexPrefix = function (nodeType) {
 module.exports = {
   'name': 'no-color-literals',
   'defaults': {
-    'allow-variable-identifiers': true,
     'allow-map-identifiers': true,
-    'allow-rgba': false
+    'allow-rgba': false,
+    'allow-variable-identifiers': true
   },
   'detect': function (ast, parser) {
     var result = [],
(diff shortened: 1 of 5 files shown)


DanPurdy:
Due to the extra features contained in this we should probably hold it for 1.7.0

There's a few fixes we could roll out already as 1.6.3

lib/rules/no-color-literals.js
```diff
@@ -8,14 +8,69 @@ var helpers = require('../helpers'),
...
+};
+
+/**
+ * Checks the see if the node type is a hex value if so return the correct prefix
+ *
+ * @param {String} nodeType - The node type identifier
+ * @returns {String} Either a '#' or an empty string
+ */
+var checkHexPrefix = function (nodeType) {
+  var prefix = nodeType === 'color' ? '#' : '';
```
bgriffith: Could be simplified with;
```js
return nodeType === 'color' ? '#' : '';
```

PR #648, opened 2016-04-26
Fix issue with placeholder name format with interpolation

### LINKED ISSUE #625: Placeholder name format doesn't ignore variables
Current the `placeholder-name-format` rule will incorrectly flag up variables.
```scss
$var: 'foo';

%#{$var} {
  font-size: 18px;
}
```

Will incorrectly return `Placeholder %#{$var} should be written in lowercase with hyphens`.

Sass 1.5.1, develop branch

DanPurdy:
Fixes #625 

Makes sure that the placeholder name is an ident rather than interpolation or a variable etc.

`DCO 1.1 Signed-off-by: Dan Purdy <dan@danpurdy.co.uk>`

### DIFF
--- a/lib/rules/placeholder-name-format.js
+++ b/lib/rules/placeholder-name-format.js
@@ -18,60 +18,60 @@ module.exports = {
           strippedName,
           violationMessage = false;
 
-      strippedName = name;
+      if (node.first().is('ident')) {
+        strippedName = name;
+        if (parser.options['allow-leading-underscore'] && name[0] === '_') {
+          strippedName = name.slice(1);
+        }
 
-      if (parser.options['allow-leading-underscore'] && name[0] === '_') {
-        strippedName = name.slice(1);
-      }
+        switch (parser.options.convention) {
+        case 'hyphenatedlowercase':
+          if (!helpers.isHyphenatedLowercase(strippedName)) {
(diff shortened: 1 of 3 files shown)


lib/rules/placeholder-name-format.js
```diff
@@ -18,60 +18,60 @@ module.exports = {
...
-        break;
-      default:
-        if (!(new RegExp(parser.options.convention).test(strippedName))) {
-          violationMessage = 'Placeholder \'%' + name + '\' should match regular expression /' + parser.options.convention + '/';
 
-          // convention-message overrides violationMessage
-          if (parser.options['convention-explanation']) {
-            violationMessage = parser.options['convention-explanation'];
+        switch (parser.options.convention) {
+        case 'hyphenatedlowercase':
```
bgriffith: Appears to be some weird indentation going on

DanPurdy: Nope, It's how eslint is set.. Changing to how we would write it gives indentation warning, also travis is already doing the linting tests & you'll wanna do the PR with ?w=1 as i only added 2 lines..

bgriffith: It's still weird haha

Issue #658, opened 2016-04-28
Return empty json array when Sass-Lint runs successfully but finds no issues.


mikesive:

DanPurdy:
We can definitely look to do this, unfortunately your PR didn't cater for the majority of users who don't use output files for their results. 

To implement you'd need to make sure the user has chosen to output the results to a file first and then check if the results are empty or not. 

Also with the different for matters we use it couldn't just be JSON format if for example the user has html output..

DanPurdy:
Also there were specific steps taken to silence the output on the event of no warnings/errors before https://github.com/sasstools/sass-lint/issues/141 so I'm not sure whether this would contradict that. @sasstools/owners thoughts?

mikesive:
I just have a specific use case where I need it to write out an empty JSON array to the ouput file. What is the exit code if the linter actually fails to run as opposed to if it finds errors?

tommycoppers:
I was looking into this as well.  I was going to submit a PR  that would remove the `if` block in the `outputResults()` method (https://github.com/sasstools/sass-lint/blob/fbaf73825dfe57bc6827d1d4bd79bba513fd5934/index.js#L264).  Then I came across issue https://github.com/sasstools/sass-lint/issues/141 .  It wouldn't quite reverse that PR completely, but it would be slightly contradictory.  

I just spent a fair amount of time trying to debug why ignoring a file was still showing a warning in the output, before I realized that the output was never being updated.  It wasn't being updated because all the warnings/errors were addressed.  Some sort of feedback that there were 0 errors would have prevented my witch hunt.  

I'd love to contribute, if you all believe consistent outputting would be a worthwhile feature.  I definitely believe it is.

DanPurdy:
@tommycoppers that's an interesting occurrence, I can see both sides of the argument. Maybe it's a case that output files should still be written but the console should remain silent on no errors. This wouldn't be hard to implement at all in the area you highlighted, I'd be happy to get PR's for this if you feel it's the way to go or we can discuss a better approach?

Thanks

PR #659, opened 2016-04-28
return empty json if no results #658

mikesive:


DanPurdy:
Sorry but you are writing a file here every time even if a user hasn't specified to write their results to a file in the first place. We definitely couldn't adopt this. In its current form.

PR #663, opened 2016-04-29
Add pseudo-element rule

### LINKED ISSUE #662: Colons for pseudo-classes and pseudo-elements
Copied verbatim from https://github.com/brigade/scss-lint/issues/592:

---

CSS3 introduces a distinction between using single or double colons for pseudo-classes and pseudo-elements. Per MDN:

> Sometimes you will see double colons (::) instead of just one (:). This is part of CSS3 and an attempt to distinguish between pseudo-classes and pseudo-elements. Most browsers support both values.

It’d be nice to have a linter to check that:

* Pseudo-**classes** have **one** colon (`:`)
* Pseudo-**elements** have **two** colons (`::`)

---

https://github.com/brigade/scss-lint/pull/634/files

gtramontina:
Resolves #662 

I've followed the same idea as https://github.com/brigade/scss-lint/pull/634/files. If the rule `pseudo-element` is enabled:
* **double** colons are enforced for pseudo-**elements**
* **single** colon is enforced for pseudo-**classes**.

_**NOTE:** I'd call this a WIP, as it does not consider vendor-prefixed pseudo elements and classes. Or should we take this as separate issue?_

---

`DCO 1.1 Signed-off-by: Guilherme Johansson Tramontina <guilherme.tramontina@gmail.com>`

### DIFF
--- a/lib/rules/pseudo-element.js
+++ b/lib/rules/pseudo-element.js
@@ -76,7 +76,7 @@ module.exports = {
           'ruleId': parser.rule.name,
           'line': node.start.line,
           'column': node.start.column,
-          'message': 'Pseudo-classes must start with double colons',
+          'message': 'Pseudo-classes must start with single colon',
           'severity': parser.severity
         });
       }
--- a/lib/rules/pseudo-element.js
+++ b/lib/rules/pseudo-element.js
@@ -1,0 +1,87 @@
+'use strict';
+
+var helpers = require('../helpers');
+
+var pseudoElements = [
+  'after',
+  'before',
+  'first-letter',
+  'first-line',
+  'selection',
+  'backdrop'
+];
+
+var pseudoClasses = [
+  'active',
+  'checked',
+  'disabled',
+  'empty',
+  'enabled',
+  'first-child',
+  'first-of-type',
+  'focus',
+  'hover',
+  'in-range',
+  'invalid',
+  'lang',
+  'last-child',
+  'last-of-type',
+  'link',
+  'not',
+  'nth-child',
+  'nth-last-child',
+  'nth-last-of-type',
+  'nth-of-type',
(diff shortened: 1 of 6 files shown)


lib/rules/pseudo-element.js
```diff
@@ -0,0 +1,87 @@
...
+      }
+    });
+
+    ast.traverseByType('pseudoElement', function (node) {
+      if (isPseudoClass(node.content[0].content)) {
+        result = helpers.addUnique(result, {
+          'ruleId': parser.rule.name,
+          'line': node.start.line,
+          'column': node.start.column,
+          'message': 'Pseudo-classes must start with double colons',
```
bgriffith: This should be `single` instead of `double` right?

bgriffith:
This is great thanks - I'll give it look over :+1:

bgriffith:
Could you add documentation for it? `/docs/rules/` and also add it to the default config `/lib/config/sass-lint.yml`

bgriffith:
Re: Vendor prefixes - Although we have a simple helper function to strip prefixes, having just had a quick look on the web, there seems to be **a lot** of vendor specific pseudo elements. 

Do you have any thoughts on what your approach would be? I think we'd be happy enough to have the first iteration vendor free, though it would tick all the boxes if they could be included.

gtramontina:
Nice... I missed that helper function. I believe that we could cover a lot of ground simply stripping them. As for the ones that are really vendor-specific (e.g `::-webkit-input-placeholder`), I can't think of anything else other than keeping a list of them somewhere, although I think this would become unmanageable…

bgriffith:
Thanks for adding the update. I agree that I could see the list becoming unmanageable pretty quickly. As a first iteration I think this is good to go :+1:

PR #671, opened 2016-05-03
Fix error with pseudo-element rule and move lists to data assets

### LINKED ISSUE #670: Pseudo-element rule doesn't like variables
The following bit of code causes the `pseudo-element` rule to derail.

```scss
    &:#{$placeholder}-placeholder {
      @content;
    }
```

This is because it's trying to use `charAt` without first checking to see if it's a string.

bgriffith:
Currently there is a bug where the rule tries to use `charAt` without first checking that it is a string.

Closes #670 and moves lists of pseudo classes and elements to data assets.


lib/rules/pseudo-element.js
```diff
@@ -1,52 +1,19 @@
...
-  'valid',
-  'visited'
-];
+var helpers = require('../helpers'),
+    yaml = require('js-yaml'),
+    fs = require('fs'),
+    path = require('path');
+
+var pseudoElements = yaml.safeLoad(
+      fs.readFileSync(path.join(__dirname, '../../data', 'pseudoElements.yml'), 'utf8')
```
DanPurdy: Do we really need these in separate files right now? They're not being used anywhere else and with such small datasets it seems unnecessary to force two file read operations on users.

bgriffith: I think they fit the bill. They're both data lists that are likely to change over time and it makes sense to have them in the `data/` directory instead of in the rule.

Having just run a quick `console.time` on it, they only take around ~0.1ms to 0.2ms each so I don't personally feel it's too expensive.

PR #681, opened 2016-05-06
Add pascal case to name format rules

### LINKED ISSUE #678: Class Name Format
Add support for pascal case to Class Name Format linter.
https://github.com/sasstools/sass-lint/blob/master/docs/rules/class-name-format.md

It would be nice to support a naming convention like [SUIT CSS](https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md).

DanPurdy:
Adds pascal case format to all name format rules

closes #678 

### DIFF
--- a/docs/rules/class-name-format.md
+++ b/docs/rules/class-name-format.md
@@ -121,6 +121,35 @@ When enabled, the following are disallowed:
 ## Example 4
 
 Settings:
+- `convention: pascalcase`
+
+When enabled, the following are allowed:
+
+```scss
+.PascalCase {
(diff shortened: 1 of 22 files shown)


DanPurdy:
I need to add the documentation and check the leading underscores for mixin name format here then it will be good to go

DanPurdy:
@benthemonkey do you wanna give this a quick look over too as I know these were yours originally

bgriffith:
This looks fine. Good catch with the violation messages and the `.` :+1:

PR #689, opened 2016-05-09
Increase Mocha timeout to 3000ms from 2000ms

bgriffith:
With the hope that AppVeyor will stop timing out and failing the tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "scripts": {
     "preversion": "npm test",
     "pretest": "eslint .",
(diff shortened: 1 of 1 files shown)


DanPurdy:
When I was refactoring the tests to speed them up for this I set this timeout to 10 seconds and still the same issue. It's an appveyor issue I'm sure

bgriffith:
Oh damn. I thought as it's only on and off that perhaps it was just over 2000ms. I still suggest we give it a go. Only other thought is that we could resort to test by test timeouts for anything creating files.

Issue #672, opened 2016-05-03
Support for node 0.10.x


ianand2:
Hi
   There seems to be one usage of path.isAbsolute in lib/config.js which is not allowing this package to be used in node 0.10
https://github.com/sasstools/sass-lint/blob/master/lib/config.js#L82
It can be fixed by using something like path-is-absolute. Let me know if you would like me to send a PR with the change

Thanks

DanPurdy:
Hi @ianand2 a PR would be very welcome. We recently took the decision to no longer 'officially' support 0.10.x but happy to still see fixes for it where necessary.

ianand2:
@DanPurdy sorry for the delay, submitted the PR

DanPurdy:
No problem at all, thanks for contributing!

DanPurdy:
Fixed from #706 will be with you for sass-lint 1.8.0 which should land this week some time.

PR #690, opened 2016-05-09
Adding path absolute polyfill

ianand2:
Hi
    This PR is for: https://github.com/sasstools/sass-lint/issues/672
I'm not sure if there is any testing you would like to see here. Travis already seems to be testing it on node 0.10, though when testing, the fs-extra package is patching the path module with the isAbsolute function, https://github.com/jprichardson/node-fs-extra/blob/9ed01a7bc016731798ebb2eee9b5933860d46811/lib/ensure/symlink-paths.js#L3, so this was not caught in tests.

DCO 1.1 Signed-off-by: Anand <ianand2@gmail.com>

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -3,7 +3,8 @@
 var yaml = require('js-yaml'),
     fs = require('fs'),
     path = require('path'),
-    merge = require('merge');
+    merge = require('merge'),
+    pathIsAbsolute = require('path-is-absolute');
 
 var cacheConfig = {},
     cacheEnabled = false;
@@ -79,7 +80,7 @@ module.exports = function (options, configPath) {
       configPath = findFile(false, '.sass-lint.yml');
     }
   }
-  else if (!path.isAbsolute(configPath)) {
+  else if (!pathIsAbsolute(configPath)) {
     configPath = path.resolve(process.cwd(), configPath);
   }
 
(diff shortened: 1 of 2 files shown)


dcai-medallia:
+1

DanPurdy:
@dcai-medallia Please don't put +1 comments in PR's or issues, you can use the Github reactions to add your support but +1's just clutter the task.

DanPurdy:
@ianand2 I'll close this PR and use #706 instead I just added some appveyor and travis changes for node 0.10.x on top of your changes thanks again 👍

ianand2:
@DanPurdy regarding the fs-extra comment, it was polyfilling path.isAbsolute when running the tests. I think the sequence of loading the libraries is such that the polyfill happens before sass-lint starts when testing, so this error was not caught as part of the test suite. This is still necessary to be able to run it in node 0.10

DanPurdy:
@ianand2 So we haven't been testing against 0.10 for a few weeks now due to the random issues we were having that seemed related to our use of child processes in our tests and the spurious error codes they would sometimes throw out when we were testing for a specific error code, your change actually seems to have resolved a lot of those so it seems like we were just looking in the wrong place!

I've re implemented the 0.10 tests for Travis and Appveyor. On appveyor it's set as an acceptable failure as there are two issues that we know aren't issues but we'll make sure all of our other tests pass.

Out of interest is there any reason why you haven't / can't upgrade your node version?

ianand2:
Fixing bugs by accident, thats always nice :) 
We are in the process of migrating, we just haven't had any blocker issues with 0.10 so far for it to be a high priority.

DanPurdy:
Makes sense, thanks! 👍

PR #691, opened 2016-05-09
Allow multiple input sources

### LINKED ISSUE #668: Allow configuring 'include' option as an array in sass-lint.yml
We have our SCSS files in two separate folders, one for global styles and another where it is co-located with Angular components. From what I've seen in projects at two different companies this is a pretty typical file layout. In order to configure sass-lint to only see the actual paths we need, the only possibility at the moment is to specify `include: '**/*.scss'` and add *everything else* to the `ignore` list - but that is counter-intuitive and not really best practice since it's prone to breaking if anything in the project structure changes.

IMHO this asymmetry between `include` and `ignore` should be eliminated. `include` should be specifiable as an array just like `ignore`, thus significantly reducing the configuration complexity (because `ignore` can in many cases simply be ignored). ;)

DanPurdy:
Allows you to add an array of globs/patterns to the include property of sass-iints config and also specify multiple patterns/paths from the CLI

You can still pass a string of one pattern to the files property but it's my feeling that this should have been removed when ignores were added as it creates an alternative way of including files. Something we should remove in 2.0 i believe.

Happy for feedback on any of the documentation changes here or on the approach.

Closes #668

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,12 @@ A Node-only Sass linter for both `sass` and `scss` syntax!
 ## Install
 You can get `sass-lint` from [NPM](https://www.npmjs.com/package/sass-lint):
 
+Install globally
+```
+npm install -g sass-lint
+```
+
+To save to a project as a dev dependency
 ```
 npm install sass-lint --save-dev
 ```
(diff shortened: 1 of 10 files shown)


DanPurdy:
Coveralls is lying, im not sure how it can say that some of that isn't being tested but oh well..

DanPurdy:
Will need some time to debug this one, it's really odd..

bgriffith:
This is looking good and will hopefully clean up some confusion around our config :+1:

PR #706, opened 2016-05-12
Absolute Path polyfill

DanPurdy:
This replaces the PR in #690 just so I could add node 0.10 in the test matrices again and add a slight tidy up.

Appveyor 0.10 is set to allow failures as it for some reason can't read the help text or version number that we generate, included screenshot proves that this isn't the case though.

screen shot 2016-05-12 at 13 08 05

We'll just have to be vigilant with the 0.10 tests and make sure only 2 failures get through though but it's worth keeping 0.10 support at least until it drops off the LTS list in a few months time.

Thanks to @ianand2 too for the original fix here.

<DCO 1.1 Signed-off-by: Dan Purdy dan@danpurdy.co.uk>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,9 @@ node_js:
   - '5'
   - node
   - iojs
+matrix:
+  allow_failures:
+    - node_js: "0.10"
 before_script: npm link
 after_success: npm run coveralls
 notifications:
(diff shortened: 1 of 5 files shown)


bgriffith:
:shipit:

PR #709, opened 2016-05-12
Non standard property prefixes

### LINKED ISSUE #702: no-vendor-prefixes : Throws error on Non-Standard Rule
When enforcing the [no-vendor-prefixes](https://github.com/sasstools/sass-lint/blob/master/docs/rules/no-vendor-prefixes.md) rule, it throws errors due to non-standard rules, as utilized in [normalize.css](https://necolas.github.io/normalize.css/4.1.1/normalize.css):

/* Remove inner padding and border in Firefox 4+.*/
`button::-moz-focus-inner,
input::-moz-focus-inner {
    border: 0;
    padding: 0;
}`

/* Fix the cursor style for Chrome's increment/decrement buttons. For certain font-size values of the input, it causes the cursor style of the decrement button to change from default` to text.*/

`input[type="number"]::-webkit-inner-spin-button,
input[type="number"]::-webkit-outer-spin-button {
    height: auto;
}`

> error  Vendor prefixes should not be used  no-vendor-prefixes

Yes, this is a vendor prefix, but it is for a non-standard rule, and is targeting Specific Browser Sudo Elements

I'm tempted to exclude our normalize.scss file from linting, but I would prefer not to.

Could there be an option, added to the rule, that would allow the user to list non-standard css rules that would then be ignored by the linter?

"gulp-sass": "2.0.4"
"gulp-sass-lint": "1.7.0"

DanPurdy:
Added the `ignore-non-standard` option to the `no-vendor-prefixes` rule. This allows you to blanket ignore the vendor prefixes on any non standard properties as defined by our definitive list. I'd like to discuss if this should also apply for pseudo elements, as far as I can see the only pseudo elements that require vendor prefixes are 'experimental' anyway so there's no need to do the extra check here

### IMPORTANT
* I've chosen to leave this new option as false to keep it functioning exactly as it had been before this change but I'm of the opinion it should probably default to true, open to suggestions here.

* I added the property 'keyframes' to the property list, this could be controversial but it seems that when prefixed Gonzales ignores the '@' part of the ident and even in some instances of the use `@keyframes` it does the same so I've added this property to ensure we can lint against it. What are peoples opinions here too? If we don't at least include it as a special case in the rule then `@-webkit-keyframes` would be considered a non-standard property.

* Updated the strip prefix helper method to cater for the above inconsistency too as originally it would have stripped the `@` had it been present. This sort of future proofs this method for any other similarly structured properties.

closes #702 

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -223,9 +223,9 @@ helpers.stripQuotes = function (str) {
  * @returns {string} The string without vendor prefixes
  */
 helpers.stripPrefix = function (str) {
-  var modPropertyArr = str.split('-');
-  var modProperty = '';
-  var prefLength = modPropertyArr[2] === 'osx' ? 2 : 1;
+  var modPropertyArr = str.split('-'),
+      modProperty = '',
+      prefLength = modPropertyArr[2] === 'osx' ? 2 : 1;
 
   modPropertyArr.splice(1, prefLength);
 
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -223,14 +223,20 @@ helpers.stripQuotes = function (str) {
  * @returns {string} The string without vendor prefixes
  */
 helpers.stripPrefix = function (str) {
-  var modProperty = str.slice(1),
-      prefixLength = modProperty.indexOf('-');
+  var modPropertyArr = str.split('-');
+  var modProperty = '';
+  var prefLength = modPropertyArr[2] === 'osx' ? 2 : 1;
 
-  // check for the -osx- vendor prefix too as well as the original -moz- etc
-  if (modProperty.slice(prefixLength + 1).slice(0, 3) === 'osx') {
-    return modProperty.slice(prefixLength + 5);
-  }
-  return modProperty.slice(prefixLength + 1);
+  modPropertyArr.splice(1, prefLength);
+
+  modPropertyArr.forEach(function (item, index) {
+    modProperty = modProperty + item;
+    if (index > 0 && index < modPropertyArr.length - 1) {
+      modProperty = modProperty + '-';
+    }
+  });
+
+  return modProperty;
 };
 
 /**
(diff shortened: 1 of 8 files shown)


lib/helpers.js
```diff
@@ -223,14 +223,20 @@ helpers.stripQuotes = function (str) {
  * @returns {string} The string without vendor prefixes
  */
 helpers.stripPrefix = function (str) {
-  var modProperty = str.slice(1),
-      prefixLength = modProperty.indexOf('-');
+  var modPropertyArr = str.split('-');
```
bgriffith: Should stick to the general way of declaring vars in sasslint, as in:

```js
 var modPropertyArr = str.split('-'),
     modProperty = '',
     prefLength = modPropertyArr[2] === 'osx' ? 2 : 1;
```

lib/rules/no-vendor-prefixes.js
```diff
@@ -1,13 +1,30 @@
 'use strict';
 
 var helpers = require('../helpers');
+var yaml = require('js-yaml');
```
bgriffith: Same as previous comment

bgriffith:
I think leaving it as false by default is ok for now as I would expect the rule to highlight every use of a vendor prefix.

Also ok with the addition of keyframes in there.

Is `osx` the only instance of a double prefix? Might be worth while turning the that into an array in case there are more instances.

DanPurdy:
It's all I could see, I'll leave it as a string for now but it's easy to update in the future if necessary as this was just a refactor of that helper method.

Romainpetit:
Documentation is missing an example :
https://github.com/sasstools/sass-lint/blob/master/docs/rules/no-vendor-prefixes.md#ignore-non-standard

I intuitively used this, not working :
```
  no-vendor-prefixes:
    - 1
    -
      ignore-non-standard: 1
```

colin-marshall:
@Romainpetit this worked for me:

```
no-vendor-prefixes:
    - 1
    - ignore-non-standard: true
```

DanPurdy:
It's under the options heading https://github.com/sasstools/sass-lint/blob/master/docs/rules/no-vendor-prefixes.md#options

Though this comment is like 3 months old so maybe it wasn't before.. Can't remember!

PR #722, opened 2016-05-18
Move back to gonzales-pe and update version

DanPurdy:
Gonzales-pe is updating again and we now have write access so we can drop the fork for the foreseeable future.

### DIFF
--- a/lib/groot.js
+++ b/lib/groot.js
@@ -3,7 +3,7 @@
 //////////////////////////////
 'use strict';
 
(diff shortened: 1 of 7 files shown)


DanPurdy:
Looks like more stuff to fix 😭

DanPurdy:
Going to use this as an ongoing test PR for the new gonzales versions, it will only be merged once 1.8.0 is ready to go. Also don't want to specify version ranges for gonzales as it bit us in the behind before.

Issue #611, opened 2016-04-15
Inaccurate Indentation Failures


allenhartwig:
Lint fails on indentation with odd results. Eg: "Indentation of 0..." (when the line does have indentation) "..., expected 2" (what it has)

**What version of Sass-Lint are you using?**
1.5.1

**Please show your full configuration:**
sass-lint.yaml
```yml
rules:
  indentation:
    - 2
```

**What did you do? Please include the actual source code causing the issue.**
`sass-lint -c sass-lint.yaml test.sass -v -q -f compact`
test.sass
```css
table
  width: 100%

  a
    &:link
      text-decoration: none

    &:visited
      text-decoration: none

    &:active
      text-decoration: underline
```
**What did you expect to happen?**
Lint passes

**What actually happened? Please include the actual, raw output from ESLint.**
Lint fails with 
```bash
test.sass: line 4, col 1, Error - Indentation of 0, expected 2 (indentation)
test.sass: line 5, col 1, Error - Indentation of 0, expected 2 (indentation)
test.sass: line 6, col 1, Error - Indentation of 0, expected 4 (indentation)
test.sass: line 9, col 1, Error - Indentation of 0, expected 4 (indentation)
test.sass: line 12, col 1, Error - Indentation of 0, expected 4 (indentation)
```

DanPurdy:
There's quite a few issues surrounding this already yes. Sass syntax is notorious for it as well as using crlf line endings or hard tabs.. The other issues have a more in depth description but unfortunately until we can get 1.6 out and investigate this with the latest version of the AST we use we can't fix it, out hands are tied.. I recommend just disabling it for now unfortunately if you match any of the criteria I mention above. :sob:

allenhartwig:
I'm using LF and soft tabs (2 spaces). Should I still be seeing this issue?

DanPurdy:
You're using sass syntax not scss right? There's still issues with scss in certain places but it's definitely more stable than sass syntax purely down to the differences the AST we're using handles those formats.

allenhartwig:
Yes, sass syntax.

Are there any other known issues right now, or is it just indentation?

Marc-Boisvert-Dupras:
I'd just like to add I'm seeing this as well (I guess that's to be expected at the moment) using Sass syntax.

Kathrynwatts:
[...]

PR #725, opened 2016-05-20
Update indentation to support tabs and fix bugs [SCSS | Sass]

### LINKED ISSUE #382: Indentation rule - Incorrect mixed spaces / tabs warnings
So this is two issues really, first raised in #373

sass-lint @ 1.3.2
node @ any version

The error being thrown in #373 is because it's checking for a non [existent node content](https://github.com/sasstools/sass-lint/blob/develop/lib/rules/indentation.js#L57) block when files are using `CRLF` line endings.

This seems to actually hide the real issue though which is an issue where `CRLF` endings are throwing false mixed spaces and tabs warnings. On first look this is due to the regex for whitespace matching new line characters when dealing with `CRLF` as we're manually slicing the content array and only taking into account the length of `\n` rather than `\r\n` as well. 

Use the file uploaded in #373 and make sure your editorconfig/IDE doesn't auto format the line endings and you'll be able to test this.

### LINKED ISSUE #426: Unexpected indenting enforcement on multiline rules
In one of our projects we have the following Sass for hi-dpi images:
```sass
//Image Manangement
@mixin hidpi-background-image($filename, $background-size: 'mixed', $extension: png) {
  background-image: url('../images/#{$filename}.#{$extension}');
  @if ($background-size != 'mixed') {
    background-size: $background-size;
  }
  @media (min--moz-device-pixel-ratio: 1.3),
    (-o-min-device-pixel-ratio: 2.6/2),
    (-webkit-min-device-pixel-ratio: 1.3),
    (min-device-pixel-ratio: 1.3),
    (min-resolution: 1.3dppx) {
      background-image: url('../images/#{$filename}@2x.#{$extension}');
    }
}
```

**sass-lint@1.3.3** (via **gulp-sass-lint**) throws the following warnings:
```
frontend/static-src/styles/_theme-helpers.scss
  61:11  warning  Vendor prefixes should not be used  no-vendor-prefixes
  62:3   warning  Indentation of 2, expected 0        indentation
  63:3   warning  Indentation of 2, expected -2       indentation
  63:4   warning  Vendor prefixes should not be used  no-vendor-prefixes
  64:3   warning  Indentation of 2, expected -4       indentation
  65:3   warning  Indentation of 2, expected -6       indentation
  66:5   warning  Indentation of 4, expected -4       indentation

✖ 7 problems (0 errors, 7 warnings)
```

Not sure why it's reporting an expected `indentation` with negative values, or how I could improve my indenting/formatting...

Is there a way to disable specific rules (like `indentation` and `no-vendor-prefixes`) for specific lines of a file while keeping it enabled elsewhere?

### LINKED ISSUE #592: Allow tab indentation
Hi,

This is more of a question than a bug. I'm wondering whether in the future the possibility to enforce tabs instead of spaces might be an option?

In this [comment](https://github.com/sasstools/sass-lint/issues/457#issuecomment-165935624) on issue #457, I see that your AST doesn't allow tabs but only spaces, but I couldn't find a dedicated issue for this problem.

Thanks

DanPurdy:
The indentation rule now supports tabs, you can specify this in your config by including the string `'tab'` instead of just a integer as you would for spaces.

```yaml
indentation:
  - 1
  -
    size: 'tab'
```

It also fixes the issue with multiline arguments to atrules. 

~~`Sass` syntax is still broken, but more consistently broken now and Ive got work in progress to sort it.~~

Lines that previously reported `mixed tabs and space` warnings also used to attempt to provide a incorrect indentation value such as ` 1 expected 2` I've removed this as it seems silly to try and lint something that is clearly an error and the mixed warning will suffice to show there's an issue with that line. 

The first space character encountered in the file will be taken as the de facto for the file and generate mixed warnings thereafter if a different character is encountered regardless of whether you've chosen tabs or spaces. If a tab and space are encountered on the first line of your file in a mixed fashion it will favour the space. I'm open to updating this functionality though. 

Split the test files into tabs and spaces versions for now. Both include mixed warnings so please be careful if editing these.

Fixes #426 
Fixes #382 
Closes #592 

### DIFF
--- a/docs/rules/indentation.md
+++ b/docs/rules/indentation.md
@@ -2,7 +2,7 @@
 
 Rule `indentation` will enforce an indentation size (tabs and spaces) and it will also ensure that tabs and spaces are not mixed.
 
-The mixed warnings and tabs warnings check will take into account what you have set in your config file whether it should expect to see spaces or tabs. If it encounters a tab anywhere in a file when your rule config doesn't specify tabs it will flag a lint warning, Similarly for any whitespace using spaces when tabs are specified. Obviously spaces between properties and values etc are ignored.
+The mixed spaces and tabs warnings check will take into account what you have set in your config file whether it should expect to see spaces or tabs. If it encounters a tab anywhere in a file when your rule config doesn't specify tabs it will flag a lint warning, Similarly for any whitespace using spaces when tabs are specified. Obviously spaces between properties and values etc are ignored.
 
 ## Options
 
--- a/docs/rules/indentation.md
+++ b/docs/rules/indentation.md
@@ -2,7 +2,7 @@
 
 Rule `indentation` will enforce an indentation size (tabs and spaces) and it will also ensure that tabs and spaces are not mixed.
 
-The mixed warnings and tabs warnings check will always choose the first tab or space that it comes across as the de facto standard for the file it's linting regardless of the options you have in your config. If a tab and a space both appear in the first indented property/block in your file the space will be preferred over the tab. This is purely to make sure you get accurate mixed spaces and tabs warnings throughout the file and any indented elements can be linted correctly.
+The mixed warnings and tabs warnings check will take into account what you have set in your config file whether it should expect to see spaces or tabs. If it encounters a tab anywhere in a file when your rule config doesn't specify tabs it will flag a lint warning, Similarly for any whitespace using spaces when tabs are specified. Obviously spaces between properties and values etc are ignored.
 
 ## Options
 
--- a/docs/rules/indentation.md
+++ b/docs/rules/indentation.md
@@ -1,10 +1,12 @@
 # Indentation
 
-Rule `indentation` will enforce an indentation size (in spaces) and ensure that tabs and spaces are not mixed.
+Rule `indentation` will enforce an indentation size (tabs and spaces) and it will also ensure that tabs and spaces are not mixed.
+
+The mixed warnings and tabs warnings check will always choose the first tab or space that it comes across as the de facto standard for the file it's linting regardless of the options you have in your config. If a tab and a space both appear in the first indented property/block in your file the space will be preferred over the tab. This is purely to make sure you get accurate mixed spaces and tabs warnings throughout the file and any indented elements can be linted correctly.
 
 ## Options
 
-* `size`: `number` (defaults to `2`)
+* `size`: `number` or `'tab'` (defaults to `2` spaces)
 
 ## Examples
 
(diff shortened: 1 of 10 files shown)


tests/sass/indentation/indentation-tabs.sass
```diff
@@ -0,0 +1,98 @@
+$foo: ('foo': 'bar', 'bar': ('foo': 'bze'))
```
DanPurdy: The Sass test files can be ignored for now. They will be updated in the Sass syntax PR for indentation once it is done.

DanPurdy:
Updated it to work with `Sass` syntax too. 

Also Fixes #611

docs/rules/indentation.md
```diff
@@ -1,10 +1,12 @@
 # Indentation
 
-Rule `indentation` will enforce an indentation size (in spaces) and ensure that tabs and spaces are not mixed.
+Rule `indentation` will enforce an indentation size (tabs and spaces) and it will also ensure that tabs and spaces are not mixed.
+
+The mixed warnings and tabs warnings check will take into account what you have set in your config file whether it should expect to see spaces or tabs. If it encounters a tab anywhere in a file when your rule config doesn't specify tabs it will flag a lint warning, Similarly for any whitespace using spaces when tabs are specified. Obviously spaces between properties and values etc are ignored.
```
bgriffith: Should that be `The mixed spaces and tabs..`?

lib/rules/indentation.js
```diff
@@ -2,22 +2,63 @@
...
 var helpers = require('../helpers');
 
+/**
+ * Check if a node is a newline character or not
+ *
+ * @param {Object} node - The node to test
+ * @returns {boolean} Whether the node is a newline or not
+ */
+var isNewLine = function (node) {
+  // using type === instead of is just in case node happens to be a string
```
bgriffith: if node was a `string` would it not have a `type` property anyway?

lib/rules/indentation.js
```diff
@@ -2,22 +2,63 @@
 
 var helpers = require('../helpers');
 
+/**
+ * Check if a node is a newline character or not
+ *
+ * @param {Object} node - The node to test
+ * @returns {boolean} Whether the node is a newline or not
+ */
+var isNewLine = function (node) {
```
bgriffith: This should probably be moved into the helpers file as it's pretty generic and we may need it again.

lib/rules/indentation.js
DanPurdy: If you pass it a node though that is a string then `node` will be truthy but instead of `node.type` being undefined it would throw a type error that `node.is is not a function` I think it's just safer overall when dealing with these nodes as we're blindly passing the next node in the sequence to it rather than check if it's a space first, which is what this function is made to do.

So yeah, I'd rather keep these this way.

PR #764, opened 2016-06-24
Disable linters temp

macdonaldr93:
Try this again.


macdonaldr93:
So sorry, was trying to do something else.

PR #767, opened 2016-06-28
Add backdrop-filter to properties list

controversial:
Added [`backdrop-filter`](https://webkit.org/blog/3632/introducing-backdrop-filters/) as a known property. This is used fairly widely, most notably on [some pages of Apple's website](http://www.apple.com/watch/). It's available (with `-webkit-`) on safari (desktop and mobile), and is available without prefix in recent Chrome Canary builds.

Although it's only widely available with prefix, it still shows as a misspelled property in configs that have vendor prefix warnings disabled.

No
**Have you written tests?**
Nope. If this is the kind of thing that needs tests, let me know, and I'll gladly write one
Is documentation relevant for this?
No open issues

`<DCO 1.1 Signed-off-by: Luke Deen Taylor luke@deentaylor.com>`

### DIFF
--- a/data/properties.yml
+++ b/data/properties.yml
@@ -19,6 +19,7 @@ app-region
 appearance
 aspect-ratio
 azimuth
+backdrop-filter
 backface-visibility
 background
 background-attachment


DanPurdy:
@controversial I'm not sure about this, it's currently listed as a non official property and under an editors draft, with usage only possible with a vendor prefix. Can you not just add it to the extra properties option within the rule instead?

https://github.com/sasstools/sass-lint/blob/develop/docs/rules/no-misspelled-properties.md

@sasstools/sass-lint-contributors any opinions?

controversial:
Yeah, I guess I could. While it's not the end of the world if this doesn't get merged, what's the argument for not merging it? The way I see it,
- It's not similar enough to any other properties that it could raise false positives with spelling, I see only 
- It's already available in Chrome without prefix, it's just hidden behind a flag, which seems to indicate that it's on the path to being supported.
- It's been used in production by Apple and a few others for more than a year.

So... why *not* merge it? It seems to me like there are only advantages, no disadvantages.

Issue #775, opened 2016-07-02
Feature: Detecting nbsp (non-breaking-spaces) in selectors


oliver----:
What version of Sass Lint are you using?
1.8.2

We recently found out that some of our CSS rules were not applied because we had a non-breaking-space (`&#160;`) in these selectors. Especially under OSX this can happen easily using `Opt+Space` accidentally.

It would be nice, to be able to detect such errors using sass-lint.

PR #776, opened 2016-07-02
Added a rule to detect nbsp in selectors.

oliver----:

Adding a linter rule that detects nbsp in selectors.

no-invalid-whitespace -> 'Found invalid whitespace'

**Have you written tests?**
yes

no

#775

`<DCO 1.1 Signed-off-by: Oliver Buchtala oliver.buchtala@gmail.com>`

Note:
Please take a look at the implementation. ATM I traverse all nodes, looking at `selectors` and `classes`. This I did without a complete study of the AST you are using, but instead as to fulfill the tests I added.
So I'd be happy to hear your opinion, if there might be more cases, that should be considered.

### DIFF
--- a/lib/config/sass-lint.yml
+++ b/lib/config/sass-lint.yml
@@ -30,6 +30,7 @@ rules:
   no-ids: 1
   no-important: 1
   no-invalid-hex: 1
+  no-invalid-whitespace: 1
   no-mergeable-selectors: 1
   no-misspelled-properties: 1
   no-qualifying-elements: 1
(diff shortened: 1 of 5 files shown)


DanPurdy:
Thanks @oliver---- I'll have a look at this later on. The only thing I can see missing at first glance is the documentation for the rule, https://github.com/sasstools/sass-lint/tree/develop/docs/rules and also adding the rule to our default config file (probably enabled as a warning in this case) https://github.com/sasstools/sass-lint/blob/develop/lib/config/sass-lint.yml

PR #806, opened 2016-08-01
Add motion-offset and motion-path to properties.yml

### LINKED ISSUE #805: motion-offset and motion-path give no-misspelled-properties
https://developer.mozilla.org/en-US/docs/Web/CSS/motion-path

intellix:
Fixes #805

### DIFF
--- a/data/properties.yml
+++ b/data/properties.yml
@@ -369,6 +369,8 @@ min-logical-width
 min-width
 min-zoom
 mix-blend-mode
+motion-offset
+motion-path
 move-to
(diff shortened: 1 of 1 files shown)


DanPurdy:
These properties are both deprecated.

intellix:
In favour of offset-path and offset-distance it seems

DanPurdy:
Just an FYI we're updating all our property rules away from our poorly maintained property list in the next releases, you should see this fixed soon.

PR #842, opened 2016-08-25
Add max-file-line-count rule

### LINKED ISSUE #841: Rule suggestion: Max file line count
Following on from #833 it would be good to port over TSLint's max-file-line-count rule too for those who wish to enforce a limit on the size/length of their files.

DanPurdy:
Adds the rule `max-file-line-count` to enforce the length of a file

If a file exceeds the default / user specified line count the following message will be provided:
`This file has 301 lines, which exceeds the maximum of 300 lines allowed.`

closes #841 


docs/rules/max-file-line-count.md
```diff
@@ -0,0 +1,27 @@
+# Max File Line Count
+
+Rule `max-file-line-count` will enforce that a files length doesn't exceed a certain number of lines
```
bgriffith: Typo - `file's`

lib/rules/max-file-line-count.js
```diff
@@ -0,0 +1,48 @@
...
+  return last ? getLastNode(last) : node;
+};
+
+module.exports = {
+  'name': 'max-file-line-count',
+  'defaults': {
+    length: 300
+  },
+  'detect': function (ast, parser) {
+    var result = [];
```
bgriffith: This could be simplified as all we need is the `end.line` of the `stylesheet` node, which will always be `ast.end.line`

bgriffith:
```js
...  
  'detect': function (ast, parser) {
    var result = [];

    if (ast.end.line > parser.options.length) {
        'ruleId': parser.rule.name,
        'line': ast.end.line,
        'column': 0,
        'message': 'This file has ' + ast.end.line + ' lines, which exceeds the maximum of ' + parser.options.length + ' lines allowed.',
        'severity': parser.severity
      });
    }

    return result;
  }
...
```

Issue #813, opened 2016-08-04
no-url-protocols doing more than checking protocols


danwaz:
I feel like this rule is overloaded. A user should be able to check for protocols separately from domains.  I'd recommend adding configuration that could enable/disable the two parts of this rule or creating a separate rule that handles the domain check.

sass-lint@1.8.2

DanPurdy:
HI @danwaz I completely agree with you. Our aim has always been for small single purpose rules and therefore I think your suggestion would perfectly suit this mindset. This would present a breaking change though so I propose maybe the following. 

1. We keep the rule working as is but we add a flag such as 'enforce-domains` too which would be on by default.
1. We add the no-url-domains rule to mimic that functionality also and enable it in our default config
1. We document that as of v2 of sass-lint the option in 1 would be removed and `no-url-protocols` would only check for url protocols and no longer check domains.

The above should give a smooth transition to existing users of the rule while also not breaking anything for those not keeping up with our changelogs. Would that sound fair?

Also on a side note we've all been a little busy which is why not much activity has happened here recently so as usual we'd be happy to see PR's for any of this from anyone interested.

danwaz:
Thanks @DanPurdy.  This sounds like a solid path forward.  I don't mind taking on the work and getting this into PR.  Also a bit busy on my end, but I'll see what I can do.

PR #846, opened 2016-08-25
Updates no-url-protocols rule and adds additional no-domains rule

danwaz:
Addresses issue #813 

This pull request addresses the conflation of the no-url-protocols and no-domains rules.  In addition, it also adds a configuration option to allow for protocol-relative-urls (i.e. `//some-domain.com/foo.png`) within the no-url-protocols rule. 

As discussed in issue #813, this new configuration should not add any breaking changes if users decide to update to this version.  All additional configurations are set to `false` by default.  Also to keep in mind, when the project updates to v2, all legacy no-domain logic should be removed from the no-url-protocols rule.

Unit tests and documentation have been updated to reflect the new rule and additional configuration.

`<DCO 1.1 Signed-off-by: Dan Wasilewski heydanwaz@gmail.com>`


DanPurdy:
Hey @danwaz would you mind updating the branch again I just had to finish merging the release branch in. I can review it for you after that.

Thanks for the PR! :+1:

danwaz:
@DanPurdy No problem! Updating now.

lib/rules/no-url-protocols.js
```diff
@@ -2,30 +2,34 @@
...
-  return str.substring(1, str.length - 1);
-};
+var isUrlRegex = /^(https?:)?\/\//,
+    protocolRelativeRegex = /^(https?:)\/\//;
 
 module.exports = {
   'name': 'no-url-protocols',
-  'defaults': {},
+  'defaults': {
+    'protocol-relative-urls': false
```
DanPurdy: I think this needs a positive or negative prefix, such as `allow-protocol-relative-urls` This would reverse the logic slightly and mean you could remove the logical operators when assigning `regexSelector` and `message` below

tests/rules/no-url-protocols.js
```diff
@@ -16,8 +16,23 @@ describe('no url protocols - scss', function () {
       done();
     });
   });
+
+  it('allow protocol-relative-urls urls [protocol-relative-urls: true]', function (done) {
```
DanPurdy: Is the duplication in `-urls urls` intentional?

lib/config/sass-lint.yml
```diff
@@ -25,6 +25,7 @@ rules:
   no-css-comments: 1
   no-debug: 1
   no-disallowed-properties: 0
+  no-domains: 0
```
DanPurdy: I think we should enable this by default, even though it will lead to a duplication in error messages it may force some people to check the docs to disable the extra option you've added in `no-url-protocols`

docs/rules/no-url-protocols.md
```diff
@@ -2,9 +2,15 @@
 
 Rule `no-url-protocols` will enforce that protocols and domains are not used within urls.
 
+## Options
+
+* `protocol-relative-urls`: `true`/`false` (defaults to `false`)
+
```
DanPurdy: A small message here about the fact the `no-domains` rule could be used instead and that this option/functionality will be deprecated in sass-lint 2.0 would be good.

docs/rules/no-domains.md
```diff
@@ -0,0 +1,37 @@
+# No Domains
+
+Rule `no-domains` will enforce that domains are not used within urls.
```
DanPurdy: Only a slight thought that maybe this should be `no-url-domains` to keep it in sync with the rule it shares functionality with. I'm completely open if you prefer this though, just a thought.

DanPurdy:
Hey @danwaz, I've left a few comments for discussion, looking forward to hearing your thoughts. Appreciate the contribution 👍

lib/rules/no-url-protocols.js
danwaz: I added `allow-protocol-relative-urls` here.  I agree, much less ambiguous.

tests/rules/no-url-protocols.js
danwaz: Good eye. Classic copy/paste error. 👎

lib/config/sass-lint.yml
danwaz: Sounds good to me. I didn't know whether or not we wanted to suppress this rule until by default until 2.0.  Added as a warning by default.

docs/rules/no-url-protocols.md
danwaz: Message added.

docs/rules/no-domains.md
danwaz: I'm a fan of consistency.  I also don't mind having the two rules right next to each other in the file system. Updated.

DanPurdy:
Awesome, looking good! I'll give it a once over again in the morning (it's late here!) and merge it in if all is good. You may want to add your 'Huge' email address as a secondary to your github account too for proper attribution of those commits! 😉

danwaz:
Thanks @DanPurdy! H/t for the email suggestion.

lib/config/sass-lint.yml
```diff
@@ -25,6 +25,7 @@ rules:
   no-css-comments: 1
   no-debug: 1
   no-disallowed-properties: 0
+  no-url-domains: 1
```
DanPurdy: Could you add this in alphabetical order within this block please

DanPurdy:
Just one more change and then this is good to go!

lib/config/sass-lint.yml
danwaz: Updated.

DanPurdy:
🎉  🎉  Thanks @danwaz !

lib/rules/no-url-domains.js
```diff
@@ -0,0 +1,33 @@
...
+  'detect': function (ast, parser) {
+    var result = [];
+
+    ast.traverseByType('uri', function (uri) {
+      uri.traverse(function (item) {
+        if (item.is('string')) {
+          var stripped = helpers.stripQuotes(item.content),
+              parsedUrl = url.parse(stripped, false, true);
+
+          if (parsedUrl.host && parsedUrl.protocol !== 'data:') {
```
pmourelle: What about `//assets.example.com/some-font.ttf` ? It fails on this and I think it shouldn't.

PR #857, opened 2016-08-31
Reimplement max-warnings as a proper option

### LINKED ISSUE #856: Make "max-warnings" a proper option
The command-line tool provides the `--max-warnings` option. This is very useful for preventing a codebase from deteriorating and generating ever more warnings.

However, we can't use this feature for [vanilla-framework](https://github.com/ubuntudesign/vanilla-framework/) as we run sass-lint through [gulp-sass-lint](https://github.com/sasstools/gulp-sass-lint) in our `gulpfile.js` instead of using the CLI.

As far as I can tell there's no way to set `program.maxWarnings` through any other method. It would be most helpful to us if this could be an [option](https://github.com/sasstools/sass-lint/tree/develop/docs/options) in the same sense as `formatter` et al., so we can set it within the `.sass-lint.yml` config file as follows:

``` yaml
options:
  max-warnings: 25
```

nottrobin:

This adds `max-warnings` as a proper option, including tests and documentation. This option can be set in the config file, passed straight through to `failOnError`, or set via the command-line.

See #856 for more detail.

No.

There's a new error message:

```
MaxWarningsExceededError: Number of warnings (257) exceeds the allowed maximum of 20.
```

**Have you written tests?**

Yes. 

```
  failures
    ✓ should raise SassLintFailureError if indentation is set to error (595ms)
    ✓ should not raise SassLintFailureError if indentation is only set to warn
    ✓ should raise MaxWarningsExceededError if warnings exceed `max-warnings` setting (133ms)
    ✓ should not raise MaxWarningsExceededError if warnings do not exceed `max-warnings` setting
```

Yes, in `docs/options/max-warnings.md`.

Closes #856.

`<DCO 1.1 Signed-off-by: Robin Winslow robin@robinwinslow.co.uk>`

### DIFF
--- a/docs/sass-lint.yml
+++ b/docs/sass-lint.yml
@@ -9,6 +9,8 @@ options:
   formatter: html
   # Output file instead of logging results
   output-file: 'linters/sass-lint.html'
+  # Raise an error if more than 50 warnings are generated
+  max-warnings: 50
 # File Options
 files:
   include: 'sass/**/*.s+(a|c)ss'
--- a/tests/failures.js
+++ b/tests/failures.js
@@ -1,0 +1,90 @@
+'use strict';
+
+var lint = require('../index'),
+    assert = require('assert'),
+    exceptions = require('../lib/exceptions');
+
+describe('failures', function () {
+  it('should raise SassLintFailureError if indentation is set to error', function (done) {
+    assert.throws(
+      function () {
+        var results = lint.lintFiles('tests/sass/indentation/indentation-spaces.scss', {rules: {indentation: 2}});  // 14 errors
+        lint.failOnError(results); // Set indentation to error
+      },
+      exceptions.SassLintFailureError
+    );
+    assert.throws(
+      function () {
+        var results = lint.lintFiles('tests/sass/indentation/indentation-spaces.scss', {}, 'tests/yml/.indentation-error.yml');  // 14 errors
+        lint.failOnError(results); // Set indentation to error
+      },
+      exceptions.SassLintFailureError
+    );
+
+    done();
+  });
+
+  it('should not raise error if indentation is only set to warn', function (done) {
+    // These should produce 55 warnings and 0 errors
+    var directResults = lint.lintFiles('sass/indentation/indentation-spaces.scss', {rules: {indentation: 1}});
+    var configResults = lint.lintFiles('sass/indentation/indentation-spaces.scss', {}, 'yml/.indentation-warn.yml');
+    lint.failOnError(directResults);
+    lint.failOnError(configResults);
+
+    done();
+  });
+
+  it('should raise MaxWarningsExceededError if warnings exceed `max-warnings` setting', function (done) {
+    assert.throws(
+      function () {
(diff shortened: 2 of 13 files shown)


docs/sass-lint.yml
```diff
@@ -9,6 +9,8 @@ options:
   formatter: html
   # Output file instead of logging results
   output-file: 'linters/sass-lint.html'
+  # Raise an error if more than 10 warnings are generated
+  max-warnings: 50
```
DanPurdy: Set to 50 but the comment mentions 10

nottrobin: Good catch. Fixed.

DanPurdy:
Looking good @nottrobin, can we add a few test cases for when the user specifically sets `max-warnings` to 0 in their config file. I can see that it will pass but just belts and braces for the future.

As a side note we will need to remember to replace that `util.inherits` when we drop support for earlier versions of node as they come off of LTS.

Thanks! 👍

nottrobin:
Good idea. I'll get those new tests written in the next couple of hours.

DanPurdy:
No rush, got a few other things to finish and add in 1.10 anyway.

nottrobin:
@DanPurdy I've now added two new tests:

- should raise MaxWarningsExceededError if warnings exceed `max-warnings` of zero
- should not raise error if no warnings even if `max-warnings` is zero

tests/failures.js
```diff
@@ -0,0 +1,90 @@
...
+    var results = lint.lintFiles('sass/indentation/indentation-spaces.scss', {});  // 55 warnings
+    lint.failOnError(results, {'max-warnings': 100}); // Max 100 warnings, should succceed
+    lint.failOnError(results, {}, 'yml/.max-100-warnings.yml'); // Max 100 warnings, should succeed
+
+    done();
+  });
+
+  it('should not raise error if no warnings even if `max-warnings` is zero', function (done) {
+    var results = lint.lintFiles('sass/success.scss', {});  // no warnings
+    lint.failOnError(results, {'max-warnings': 0}); // Max 100 warnings, should succceed
```
DanPurdy: The comments are still for max 100 tests here

DanPurdy:
Thanks just the copy n pasted comments to deal with and then this lgtm for me! 

👍

tests/failures.js
nottrobin: Good point, again. I'll fix this.

nottrobin: Done

DanPurdy:
Awesome, thanks @nottrobin looking good.

DanPurdy:
Sorry @nottrobin I seemed to miss a unfortunate outcome from this to do with silencing errors and preventing unhandled exceptions with the `--no-exit` flag on the cli which lead to this change always throwing an unhandled error if a rule of severity 2 detected an issue. 

I've had to sort of half patch the old functionality on top of yours so you should still encounter exit code 1 when max-warnings fails but the error itself won't be thrown if you're specifiying `--no-exit` or `-q`.

Let me know if there's any issues here and we can work a solution out.

Issue #861, opened 2016-09-05
border-zero: A value of `0` is not allowed. `0` must be used.


richarddewit:
**What version of Sass Lint are you using?**
1.9.1

**Please include any relevant parts of your configuration**
```
border-zero:
    - 1
    - convention: 0
```

What did you do? Please include the actual source code causing the issue.
```
.element {
  border: 0;
}
```

**What did you expect to happen?**
Not raising a warning, since the actual value is the same as the expected value.

**What actually happened? Please include any error messages given to you by Sass Lint.**
(Using Atom's plugin)
```
[sass-lint] [Warning] [border-zero] A value of `0` is not allowed. `0` must be used.
```

**If you're using a IDE plugin have you tried the CLI too?**
Same output:
```
143:22  warning  A value of `0` is not allowed. `0` must be used                   border-zero
```

thijskramer:
I think this problem could be solved by changing this in [border-zero.js](https://github.com/sasstools/sass-lint/blob/develop/lib/rules/border-zero.js#L31):

**From:**
```js
if (node.content === '0' || node.content === 'none') {
  //...
}
```
**To:**
```js
if (node.content === 0 || node.content === 'none') {
  //...
}
```

Configurations with `convention: 0` won't generate false warnings this way.

richarddewit:
As I can see now, the default is set to the string `'0'`, instead of the int. When using a string, it doesn't produce a warning.

I'll pass in a PR to fix this, so that the int 0 could also be used.

richarddewit:
Btw, @thijskramer - your solution doesn't fix it, it just skips the validation. `node.content` is the value in `border: 0;` or `border: none;` which is always a string. Therefore it will not pass your if-statement because `node.content` is `'0'` (string), not `0` (int) or `'none'`

thijskramer:
Allright, well then this issue could be fixed by parsing the yml config file in a different way, because people like me are using a `.sass-lint.yml` file with the following config:
```yml
rules:
  border-zero:
    - 1
    - convention: 0
```

OR, maybe use just two equal signs to compare (I'm not sure if this would work though): 
```js
if (node.content == '0' || node.content === 'none') {
  //...
}
```

Which solution do you prefer?

richarddewit:
The one I've made the PR with :) (convert convention to string)

DanPurdy:
[...]

PR #890, opened 2016-10-17
Fix false positive in border-zero

richarddewit:
This fixes a false positive when `convention` is set to the integer `0` (as opposed to the string, which does work), by casting the `convention` to a string.

No

**Have you written tests?**
No but I've tested it manually

Not necessary

#861

`<DCO 1.1 Signed-off-by: Richard de Wit henk.exe@gmail.com>`

### DIFF
--- a/lib/rules/border-zero.js
+++ b/lib/rules/border-zero.js
@@ -29,7 +29,7 @@ module.exports = {
             var node = item.content[0];
             if (node.type === 'number' || node.type === 'ident') {
               if (node.content === '0' || node.content === 'none') {
-                if (parser.options.convention !== node.content) {
+                if (('' + parser.options.convention) !== node.content) {
                   result = helpers.addUnique(result, {
                     'ruleId': parser.rule.name,
                     'line': node.start.line,


lib/rules/border-zero.js
```diff
@@ -29,7 +29,7 @@ module.exports = {
             var node = item.content[0];
             if (node.type === 'number' || node.type === 'ident') {
               if (node.content === '0' || node.content === 'none') {
-                if (parser.options.convention !== node.content) {
```
bgriffith: Hey @richarddewit - Thanks for the PR! Just one thing - while the solution is valid, could we be a bit more descriptive and use `.toString()` instead?

lib/rules/border-zero.js
```diff
@@ -29,7 +29,7 @@ module.exports = {
...
+                if (('' + parser.options.convention) !== node.content) {
```
bgriffith: Hey @richarddewit - Thanks for the PR! Just one thing - while the solution is valid, could we be a bit more descriptive and use .toString() instead?

bgriffith:
Closing this for https://github.com/sasstools/sass-lint/pull/913. Thanks again @richarddewit :+1:

PR #893, opened 2016-10-18
Fix BOM markers causing parse issues

### LINKED ISSUE #880: UTF-8 w/ BOM + @mixin causes fatal parsing error
Demo repo [here](https://github.com/mattdistefano/sass-lint-bom-issue-demo)

Short version is that when you have a file encoded as UTF-8 w/ BOM that starts with `@mixin`, the CLI linter fails with a `1:1  error  Please check validity of the block starting from line #1  Fatal`.

UTF-8 w/out BOM works fine, as does UTF-8 w/ BOM without `@mixin` at the start. 

Happens on OS X w/ node 6.2.0 and Windows 10 w/ node 5.8.0.

The gulp-sass-lint plugin actually works fine in all scenarios.

DanPurdy:
As discussed in #880 BOM markers in files were causing issues with the way gonzales is parsing the files we send it. This change strips any BOM markers from the files text before it is parsed.

Fixes #880 

### DIFF
--- a/lib/groot.js
+++ b/lib/groot.js
@@ -5,13 +5,13 @@
 
 var gonzales = require('gonzales-pe');
 var fm = require('front-matter');
-var stripBom = require('strip-bom');
+var helpers = require('./helpers');
 
 module.exports = function (text, syntax, filename) {
   var tree;
 
   // Run `.toString()` to allow Buffers to be passed in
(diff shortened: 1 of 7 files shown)


DanPurdy:
Oh yeah, module requires ` >= node 4.0`

PR #903, opened 2016-10-22
Fix operators in variable names

### LINKED ISSUE #901: variable name format bug
The following throws a warning for variable name format, when it shouldn't (likely due to no spaces between variable name and operator):

```scss
.content-main {
  padding: $mobile-site-gutter*1.5;
}
```

DanPurdy:
If you had a variable followed by an operator i.e. `$my-var*1.5` Gonzales would report the variable name as `$my-var*1` therefore I've added a small bit of regex to split the string on these occasions which will cover the majority of problems we'd see.

fixes #901 

### DIFF
--- a/lib/rules/variable-name-format.js
+++ b/lib/rules/variable-name-format.js
@@ -16,8 +16,7 @@ module.exports = {
     ast.traverseByType('variable', function (variable) {
       var strippedName,
           violationMessage = false,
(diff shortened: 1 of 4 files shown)


bgriffith:
This appears to be fixed in the `v3.4.6` release of Gonzales, so it's likely we don't need this anymore. Could you take out your regex and try the tests?

DanPurdy:
I don't think it would harm having it though just as a belts and braces for the future?

PR #911, opened 2016-10-24
Correct cli tests to use bin

bgriffith:
Fixes an issue where the CLI tests would fail if the user had not symlinked sass-lint to the globals.

~~In progress - need to fix on Windows~~ 🚀

`<DCO 1.1 Signed-off-by: Ben Griffith gt11687@gmail.com>`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ node_js:
   - '5'
   - node
   - iojs
-before_script: npm link
 after_success: npm run coveralls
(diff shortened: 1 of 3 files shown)


DanPurdy:
You can remove https://github.com/sasstools/sass-lint/blob/develop/.travis.yml#L9 and https://github.com/sasstools/sass-lint/blob/develop/appveyor.yml#L28 as part of this.

Issue #921, opened 2016-10-31
Outdated concentric.yml


lucasjahn:
What version of Sass Lint are you using?
1.9.1

Please include any relevant parts of your configuration
I have the standard npm package using the concentric property ordering.

What did you do? Please include the actual source code causing the issue.
I updated the concentric.yml in /lib/config/property-sort-orders/ to match the in the comments highlighted list:

```
# Based on: https://github.com/brigade/scss-lint/blob/master/data/property-sort-orders/concentric.txt
```

you are now able to get a clean output if you want to use flex properties.

***solved in an open pullrequest***
#922

DanPurdy:
closed in #922

PR #922, opened 2016-10-31
updated concentric rules to match the newest property sort order list…

lucasjahn:
I updated the concentric.yml to match the newest version so we can get rid of the error messages regarding flexbox properties or background-clip.

Nope.

**Have you written tests?**
Nope.

No, because it is jut an update of the concentric.yml

#921

`<DCO 1.1 Signed-off-by: Lucas Jahn lucas@ludwig-jahn.com>`


bgriffith:
Great! Thanks a lot for the PR :+1:

Issue #268, opened 2015-10-08
New Rule: Force pseudo element/class nesting


DanPurdy:
As discussed in #256

Rule should enforce the nesting of pseudo elements/classes

esr360:
> A value of `0` is not allowed. `zero` must be used

This seems wrong to me? Setting this to `zero` should surely allow `0` instead of `none`? As far as I'm aware `border: zero` is not valid.

DanPurdy:
@esr360 sorry, what are you referring to and where?

esr360:
Sorry @DanPurdy I posted this on the wrong issue, I had another relevant issue open relating to this, I meant to post this here: https://github.com/sasstools/sass-lint/issues/861

I will open a new issue though.

DanPurdy:
It's ok no problem, just got a little confused! 👍

DanPurdy:
Hey @esr360 I've put a PR in to address your issue, feel free to add any comments there if you feel necessary. see #926

PR #926, opened 2016-11-02
Prevent invalid conventions border-zero

DanPurdy:
It was highlighted in a [recent comment](https://github.com/sasstools/sass-lint/issues/268#issuecomment-257543860) that any convention could be supplied to the border-zero rule and no validity checking would occur leading to misleading messages such as to use the convention 'zero'.

While the docs do specify the conventions it's probably a good belts and braces measure to check ourselves and default to a good choice if it seems invalid.

This PR addresses this. An extra warning will be thrown in each file a border-zero issue occurs to inform the user of the invalid content in their config file.


esr360:
Amazing, thanks so much for the speedy update.

PR #937, opened 2016-11-07
Fix cli error throws

DanPurdy:
With the introduction of the `max-warnings` option for config files in #857 an error was introduced where any errors detected by sass-lint would always throw an unhandled exception even if the `--no-exit` / `-q` option were specified.

This was unintended.

The following should now be true for all CLI instances

* If `-q` then all unhandled errors and exits should be quiet, an error code of 1 will still be present
* If max warnings fails and the CLI is 'quiet' then no errors will be thrown but an error code of 1 will be present
* If not quiet then any errors detected (i.e. rules set to 2) will throw an exception and exit the process on the first error detected.
* If not quiet and no errors are detected but the max-warnings option is exceeded then a max-warnings exception will be thrown

@nottrobin you may want to give this a once over too as it may affect your initial intentions with the PR you made. Unfortunately without a major version bump I don't think the exact implementation of yours would work, sorry that I missed this eventuality at the time.

### DIFF
--- a/bin/sass-lint.js
+++ b/bin/sass-lint.js
@@ -6,18 +6,30 @@ var program = require('commander'),
     lint = require('../index');
 
 var configPath,
-    configOptions = {};
+    config,
+    configOptions = {},
+    exitCode = 0;
 
-var detectPattern = function (pattern) {
-  var detects;
+var tooManyWarnings = function (detects, userConfig) {
+  var warningCount = lint.warningCount(detects).count;
 
-  detects = lint.lintFiles(pattern, configOptions, configPath);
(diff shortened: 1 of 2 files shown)


nottrobin:
@DanPurdy Thanks for flagging me here. Sorry for introducing this bug. It sounds like you've barely changed the actual interface from what I created - the only difference is that you're suppressing errors if `-q` is specified, as you should. Is that correct? That's fine with me.

DanPurdy:
No problem, It was actually a weak test of mine that wasn't flagging this. Exactly, haven't changed too much just had to go a bit of a roundabout way to achieve it. Just wanted to make sure you were aware in case it causes you issues 👍

PR #998, opened 2017-01-17
Don't ignore comments in no-empty-rulesets

### LINKED ISSUE #968: no-empty-rulesets does not catch rulesets containing only comments
1.10.2

With `no-empty-rulesets` enabled, rulesets containing only comments are not considered empty, and therefore do not trigger the rule. I would argue rulesets like the following are definitely empty:

```scss
.baz {
  // Here is a comment
  // These comments often contain rules that were commented out
  // width: 4px;
}
```

Does this seem like a reasonable assumption?

I worked up a commit that changes the behavior to fix this if there's any interest.

royaldark:
This commit makes the `no-empty-rulesets` rule trigger when a ruleset
contains only comments.

Closes #968.

`DCO 1.1 Signed-off-by: Joe Einertson <jeinerts@umm>`

### DIFF
--- a/lib/rules/no-empty-rulesets.js
+++ b/lib/rules/no-empty-rulesets.js
@@ -18,7 +18,9 @@ module.exports = {
       else {
         block.traverse(function (item) {
           if (!helpers.isEqual(block, item)) {
-            if (item.type !== 'space') {
+            if (item.type !== 'space' &&
+                item.type !== 'singlelineComment' &&
+                item.type !== 'multilineComment') {
               nonSpaceCount++;
             }
           }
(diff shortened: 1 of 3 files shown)


royaldark:
I thought from the discussion in #968 it would be okay since rules can now be disabled via comments. Putting it in a flag isn't much work, but I guess I don't see the use case for allowing empty blocks at all. I tried searching through issues to find relevant discussions but didn't find much.

DanPurdy:
Yeah, you're completely right here. Sorry it's taken so long to come back to this.

PR #1007, opened 2017-01-30
Throw error if loadConfig cannot find file by path

### LINKED ISSUE #1006: User is not told if given config filepath does not exist and default rules are used instead
If given a path to a file that doesn't exist, sass-lint should throw an error to let the user know it cannot find the config file. Otherwise the user will have no idea that the default config is being used and theirs is being ignored.

## Info

>What version of Sass Lint are you using?

`v1.10.2`

>Please include any relevant parts of your configuration

N/A

>What did you do? Please include the actual source code causing the issue.

I tried to load a custom yaml config file with grunt-sass-lint but it wasn't used. Tried to go into the sass-lint code and found `loadConfig()` in `lib/config-helpers.js` that was always returning an empty config for my given path.

Turns out I had given a path to a non-existent file.

>What did you expect to happen?

I expected sass-lint to tell me that it couldn't find my config file.

>What actually happened? Please include any error messages given to you by Sass Lint.

sass-lint ran with the default rules, so I got indentation warning with `size: 2` (default rule) instead of error with `size: 4` (my custom rule).

>If you're using a IDE plugin have you tried the CLI too?

Tested with CLI: all tests point to the source problem at `loadConfig()` in `lib/config-helpers.js`.

henrahmagix:
If given a path to a file that doesn't exist, sass-lint should throw an
error to let the user know it cannot find the config file. Otherwise the
user will have no idea that the default config is being used and theirs
is being ignored.

When not given a config path, continue as usual: return an empty config.

This also fixes relative path issues in the failures test suite.

---

They let the user know if the given path to a config file does not exist instead of silently returning an empty config.

Yes: 1 new `Error` thrown.

```
> confHelpers.loadConfig('notexist.yml')
Error: Cannot find configPath notexist.yml
    at [stack trace...]
```

**Have you written tests?**

Yes, and fixed the config file paths in `tests/failures.js`.

No: I think the error thrown should be the natural result of giving an incorrect filepath, rather than a "new feature".

Fixes #1006

`<DCO 1.1 Signed-off-by: Henry Blyth <blyth.henry@gmail.com>`

### DIFF
--- a/lib/config-helpers.js
+++ b/lib/config-helpers.js
@@ -56,6 +56,9 @@ var loadConfig = function (cPath) {
     if (fs.existsSync(configPath)) {
       resolvedConfig = yaml.safeLoad(fs.readFileSync(configPath, 'utf8')) || {};
     }
+    else {
+      throw new Error('Cannot find configPath ' + configPath);
+    }
   }
 
   return {
--- a/tests/config-helpers.js
+++ b/tests/config-helpers.js
@@ -78,8 +78,8 @@ describe('config-helpers', function () {
       done();
     });
 
-    it('should return an empty config when an incorrect path is provided', function (done) {
-      var testConfig = configHelpers.loadConfig('tests/yml/non-existant.yml');
+    it('should return an empty config when no path is provided', function (done) {
+      var testConfig = configHelpers.loadConfig('');
 
       assert(equal(emptyConfig, testConfig, {
         'strict': true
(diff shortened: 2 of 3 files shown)


DanPurdy:
Thanks for this PR, I'm not sure if the unhandled error throwing is really the way to go at the moment, let me run this and have a look at it.

henrahmagix:
Hi @DanPurdy, 

Thanks for taking a look. What do you think now?

Cheers,
Henry

PR #1016, opened 2017-02-12
Fix Typos

babakhani:
Just fix typo in doc


DanPurdy:
This isn't a typo, sorry but the docs were correct as is.

PR #1018, opened 2017-02-13
remove text-fill-color from properties

razvanmitre:
It is not a standard property, and autoprefixer does not treat it as such, and it does not get prefixed, thus the no-vendor-prefixes rule fails.
https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-text-fill-color

**Have you written tests?**

`<DCO 1.1 Signed-off-by: YOUR_FULL_NAME YOUR_EMAIL>`

### DIFF
--- a/data/properties.yml
+++ b/data/properties.yml
@@ -516,7 +516,6 @@ text-emphasis
 text-emphasis-color
 text-emphasis-position
 text-emphasis-style
-text-fill-color
 text-height
 text-indent
 text-justify


DanPurdy:
This is going to require a different approach im afraid. If we remove it from the file then the property spelling rule will then fail to lint it properly. 

I'm going to look at fully separating these rule lists out so we can work on them in isolation.

I'll close this for now, but know that I will take it into consideration when updating the other rules.

PR #1035, opened 2017-03-09
update quotes rule to support specific value for attribute selectors 

### LINKED ISSUE #1017: Different settings for quotes inside selector vs values ?
Hi,

First of all, thanks for the great work!

Is there a way to have single quotes for declarations/properties

```yaml
attribute-quotes: 1
quotes: 1
```

and to allow double quotes for attributes selectors?

```scss
[type="text"] {}
```

Actually, I get this message

```
Strings must use single quotes  quotes
```

Thanks a lot!

-----

What version of Sass Lint are you using? __1.10.2__

```yaml
  quotes: 1
  attribute-quotes: 1
```

```scss
[type="text"] {}
```

What did you expect to happen? __no error__

What actually happened? Please include any error messages given to you by Sass Lint.  __Strings must use single quotes__

If you're using a IDE plugin have you tried the CLI too? __Yes, both__

thierrymichel:
Closes #1017

Single quotes is a best practice.
But, for attribute selectors, it would be nice to be able to use double quotes.

I added a new `attribute` property with `single` as default value (to keep backward compatibility despite I think that `double` would have been better).

- **Are there any new warning messages?**: no
- **Have you written tests?**: yes
- **Have you included relevant documentation**: yes
- **Which issues does this resolve?**: #1017

`<DCO 1.1 Signed-off-by: Thierry Michel thmichel@gmail.com>`

### DIFF
--- a/docs/rules/quotes.md
+++ b/docs/rules/quotes.md
@@ -1,10 +1,12 @@
 # Quotes
 
 Rule `quotes` will enforce whether single quotes (`''`) or double quotes (`""`) should be used for all strings.
+You can specify a different value for attribute selectors.
 
 ## Options
 
 * `style`: `single`/`double` (defaults to `single`)
+* `attribute`: `single`/`double` (defaults to `single`)
 
 ## Examples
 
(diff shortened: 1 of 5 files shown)


thierrymichel:
Another option would be to use `quotes` only for values (`traverseByType('value')`, then check 'string' types) and use `attributes-quotes` with a new `style` property…

PR #1094, opened 2017-06-28
Selectors must be placed on single lines

johanberonius:

Each individual selector occupy only a single line.

No

**Have you written tests?**

Yes

Yes

Forgotten trailing commas are reported when single-line-per-selector is enabled.

### DIFF
--- a/tests/sass/single-line-per-selector.sass
+++ b/tests/sass/single-line-per-selector.sass
@@ -19,6 +19,11 @@
   content: 'foo'
 
 
+.foo
+.boo
+  content: 'foo'
+
+
 .foo > a,.bar
   content: 'baz'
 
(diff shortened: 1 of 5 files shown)


tests/sass/single-line-per-selector.sass
```diff
@@ -19,6 +19,11 @@
   content: 'foo'
 
 
+.foo
+.boo
+  content: 'foo'
+
+
 .foo > a,.bar
```
eddiemonge: won't this line complain?

johanberonius: Yes, two selectors on the same line still gives a warning, that has not changed in this pull request.
What is new is that the selector on line 22 that is split over two lines now gives a warning.

PR #1106, opened 2017-08-26
Handle multiline properties & update css list

### LINKED ISSUE #720: Ignore interpolated values for misspelled properties
When linting for misspellings, interpolated values within property names warn of a spelling mistake. It would be good to exclude these somehow.

For instance

``` scss
$u-layout-divider-types: (
    bottom,
    left,
    right,
    top
);

@each $type in $u-layout-divider-types {
    .u-layout-divider-#{$type} {
        border-#{$type}: 1px solid #000 !important;
    }
}
```

results in "Property `border-` appears to be spelled incorrectly"

Using v1.7.0 of sass-lint through gulp-sass

### LINKED ISSUE #805: motion-offset and motion-path give no-misspelled-properties

### LINKED ISSUE #951: Property `interpolation-mode` appears to be spelled incorrectly
I am using `gulp-sass-lint v1.3.2` with `sass-lint v1.10.2`.

For the following code:
```scss
  img {
    // Get rid of gap under images by making them display: inline-block; by default
    display: inline-block;
    vertical-align: middle;

    // Grid defaults to get images and embeds to work properly
    max-width: 100%;
    height: auto;
    -ms-interpolation-mode: bicubic;
  }
```
(from [foundation-sites / _global.scss](https://github.com/zurb/foundation-sites/blob/546a5b9f38487450df0dcf4957ba8c73bf11194e/scss/_global.scss#L156))

The following warning is throw:
```
163:5   warning  Property `interpolation-mode` appears to be spelled incorrectly  no-misspelled-properties
```

The sass-lint config is :
```yml
...
  no-misspelled-properties:
    - 1
    - extra-properties: []
...
```
(from [.sass-lint.yml](https://gist.github.com/ncoden/ec50bad6cd2f08df0780e9f356442794))

I expect this warning to not be thrown. `-ms-interpolation-mode` is an non-standard property for IE<=8 (see https://developer.mozilla.org/fr/docs/Web/CSS/Image-rendering)

**Edit**: I have the same error for `-webkit-overflow-scrolling`.

### LINKED ISSUE #1024: Rule no-misspelled-properties not working for -webkit-touch-callout
When I use the properties `-webkit-touch-callout` i got a warning.

``` css
.my-class {
  -webkit-touch-callout: none;
}
```
```
warning  Property `touch-callout` appears to be spelled incorrectly                 no-misspelled-properties
```

It seems that the rule [no-misspelled-properties](https://github.com/sasstools/sass-lint/blob/b7bcb3a4ec1a9c08d3883eb8c2acb97b6e6513c3/lib/rules/no-misspelled-properties.js#L34) remove the `-webkit-` prefix and there is no `touch-callout` in the [properties.yml](https://github.com/sasstools/sass-lint/blob/b7bcb3a4ec1a9c08d3883eb8c2acb97b6e6513c3/data/properties.yml).

A quick fix is to add `touch-callout` in the `extra-properties` option in the .sass-lint.yml file:
``` yml
 no-misspelled-properties:
    - 1
    - extra-properties: ["touch-callout"]
```

But for long term, the rule should check for `-webkit-touch-callout` and this should be in the properties.yml list because `touch-callout` is not [standard](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-touch-callout).

Using:
- Sass-lint 1.10.2
- Visual Studio Code 1.9.1
- VSCode plugin Sass Lint 0.0.3

### LINKED ISSUE #1045: CSS Grid
There are a number of [CSS Grid properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout) that are [not in `properties.yml`](https://github.com/sasstools/sass-lint/blob/develop/data/properties.yml). We should 2x check that property list with the full, final, shipped list from MDN and update accordingly

---

* 1.10.2, from CLI

* ` - no-misspelled-properties: true`

```scss
grid-gap: 1em;
grid-row-gap: 1em;
```

What did you expect to happen?

* I expected no misspelled properties as they are [valid CSS Grid properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout)

What actually happened? Please include any error messages given to you by Sass Lint.

```
error  Property `grid-gap` appears to be spelled incorrectly      no-misspelled-properties
error  Property `grid-row-gap` appears to be spelled incorrectly  no-misspelled-properties
```

### LINKED ISSUE #1101: `caret-color` property is being mistakenly considered misspelled.
**What version of Sass Lint are you using?**
0.2.x

**Please include any relevant parts of your configuration.**
```
rules:
  no-misspelled-properties: 2
```

**What did you do? Please include the actual source code causing the issue.**
```
input,
select,
textarea{
	caret-color: mix($shade, $evergreen-neue, 20%);
}
```

**What did you expect to happen?**
`caret-color` is a valid CSS property.

- [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/caret-color)
- [W3C](https://www.w3.org/TR/css-ui-3/#caret-color)

**What actually happened? Please include any error messages given to you by Sass Lint.**
In the report, I get the error: “Property `caret-color` appears to be spelled incorrectly”

**If you're using a IDE plugin have you tried the CLI too?**
I was using the CLI.

DanPurdy:
Fixes issues with the `no-misspelled-properties` rule. It currently doesn't work very well with multiline properties and only does partial string matches without trying to understand the context which in turn means it would let spelling errors through.

e.g.
```
.foo {
  colors: ... // would pass because of -moz-border-bottom-COLORS being a real property
}
```

I've also moved away from our own maintained list of properties to a more coherent, full and more importantly maintained list [known-css-properties](https://www.npmjs.com/package/known-css-properties). Any properties that aren't featured there can either be requested to be added there OR added to the extra properties setting of this rule. This will lead to a lot less frustration with our poorly maintained list.

fixes #1101 
fixes #1024
fixes #1045
fixes #951
fixes #805 
fixes #720 
closes #1062 
closes #767 

I'll update the vendor prefixes rule on top of this too and then we can removed our properties file.

### DIFF
--- a/lib/rules/no-misspelled-properties.js
+++ b/lib/rules/no-misspelled-properties.js
@@ -1,12 +1,8 @@
 'use strict';
 
-var helpers = require('../helpers'),
-    yaml = require('js-yaml'),
-    fs = require('fs'),
-    path = require('path');
-
-var properties = yaml.safeLoad(fs.readFileSync(path.join(__dirname, '../../data', 'properties.yml'), 'utf8')).split(' ');
-
+var helpers = require('../helpers');
+var properties = require('known-css-properties').all;
+var PROP_DIVIDER = '-';
 /**
  * Combine the valid property array and the array of extras into a new array
  *
@@ -18,6 +14,40 @@ var getCombinedList = function (props, extras) {
   return props.concat(extras);
 };
 
+var buildPartialProperty = function (valBlock, currentProperty, propsCounted) {
+  var propList = [];
+  var propsEncountered = propsCounted;
+  if (valBlock.contains('declaration')) {
+    valBlock.forEach('declaration', function (node) {
+      var prop = node.first('property');
+      var value = node.first('value');
+      propsEncountered++;
+
+      if (prop.first().is('ident')) {
+        if (value.contains('block')) {
+          propList = propList.concat(
+            buildPartialProperty(value.first('block'),
+              {
+                name: currentProperty.name + PROP_DIVIDER + prop.first('ident').content
+              },
+              propsEncountered
+            )
+          );
+        }
+        else {
+          propList.push({
+            name: currentProperty.name + PROP_DIVIDER + prop.first('ident').content,
+            line: prop.first().start.line,
+            col: prop.first().start.column,
+            propsEncountered: propsEncountered
+          });
+        }
+      }
+    });
+  }
+  return propList;
+};
+
 module.exports = {
   'name': 'no-misspelled-properties',
   'defaults': {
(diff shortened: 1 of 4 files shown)


lib/rules/no-misspelled-properties.js
```diff
@@ -18,38 +15,133 @@ var getCombinedList = function (props, extras) {
...
+            name: curProperty,
+            line: prop.first('ident').start.line,
+            col: prop.first('ident').start.column
+          },
+          0
+        );
+      }
+      // If we have multiline properties
+      if (fullProperties.length) {
+        fullProperties.forEach(function (constrProp) {
```
bgriffith: What does `constrProp` mean?

lib/rules/no-misspelled-properties.js
```diff
@@ -18,38 +15,133 @@ var getCombinedList = function (props, extras) {
...
+          // Add the number of propertiy declarations we've already checked here so we can skip them
```
bgriffith: Typo - `...number of property...`

lib/rules/no-misspelled-properties.js
```diff
@@ -18,38 +15,133 @@ var getCombinedList = function (props, extras) {
...
+              'line': constrProp.line,
+              'column': constrProp.col,
+              'message': 'Property `' + constrProp.name + '` appears to be spelled incorrectly',
+              'severity': parser.severity
+            });
+          }
+        });
+      }
+      else {
+        if (curProperty && curProperty.length > 0) {
```
bgriffith: Could combine the else if and just run with `...curProperty.length)`

lib/rules/no-misspelled-properties.js
DanPurdy: constructed property... I think 😄

lib/rules/no-misspelled-properties.js
DanPurdy: curProperty 'could' technically be undefined here. Would rather leave the check just in case as we don't wanna get length of undefined. Can remove the `> 0` though as that's unnecessary

lib/rules/no-misspelled-properties.js
```diff
@@ -18,38 +15,133 @@ var getCombinedList = function (props, extras) {
...
+              propsEncountered
+            )
+          );
+        }
+        else {
+          propList.push({
+            name: currentProperty.name + PROP_DIVIDER + prop.first('ident').content,
+            line: prop.first().start.line,
+            col: prop.first().start.column,
+            propsEncountered: propsEncountered
```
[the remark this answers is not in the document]
DanPurdy: Yeah... would be good. We can move up soon.

lib/rules/no-misspelled-properties.js
bgriffith: Aye, sorry the `> 0` is what I was talking about :+1:

PR #863, opened 2016-09-05
Fix Enhanced trailing semicolon to lint double semicolons

JordanDDisch:
Enhanced 'trailing-semicolon' rule to lint double semicolons.

e.g
`.foo { 
background: red;;
 }` 

**Are there any new warning messages?**
No.

**Have you written tests?**
No. I increased the asserted number of errors in the existing 'trailing-semicolon' test, after including test examples for double semicolons inside the trailing-semicolon.scss file.

**Have you included relevant documentation**
No. None was needed (please let me know if otherwise).

**Which issues does this resolve?**
#852

`<DCO 1.1 Signed-off-by: Jordan Dustin Disch jordanddisch@gmail.com>`

PR #1107, opened 2017-08-26
Feature lint double semi colons

### LINKED ISSUE #852: Lint double semi colons
Currently with trailing-semicolon on, double semi colons are not being linted.

e.g.
`.test {
  background: red;;
}`

As mentioned here in scss-lint as possible fix:
https://github.com/brigade/scss-lint/commit/0fc7ba821caf968505d3d130f60a94e832295803

Using gulp sass lint: "gulp-sass-lint": "^1.1.1"

DanPurdy:
I was unable to update a contributors branch so just created a new quick PR here. Full credit to @JordanDDisch
see #863 for details

closes #863 
Fixes #852

### DIFF
--- a/lib/rules/trailing-semicolon.js
+++ b/lib/rules/trailing-semicolon.js
@@ -13,7 +13,7 @@ module.exports = {
     var hasDoubleSemiColon = function (tree) {
 
       tree.traverseByType('declarationDelimiter', function (node, index, parent) {
-        if (parent.content[index + 1].is('declarationDelimiter')) {
+        if (parent.content[index + 1] && parent.content[index + 1].is('declarationDelimiter')) {
           result = helpers.addUnique(result, {
             'ruleId': parser.rule.name,
             'severity': parser.severity,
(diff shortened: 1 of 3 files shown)


lib/rules/trailing-semicolon.js
```diff
@@ -10,6 +10,21 @@ module.exports = {
   'detect': function (ast, parser) {
     var result = [];
 
+    var hasDoubleSemiColon = function (tree) {
+
+      tree.traverseByType('declarationDelimiter', function (node, index, parent) {
+        if (parent.content[index + 1].is('declarationDelimiter')) {
```
bgriffith: Woud check that `parent.content[index + 1]` exists before calling `is`

DanPurdy:
Just noticed that this looks wrong though, 4 failure cases added but only increments the warning count by 3.

lib/rules/trailing-semicolon.js
DanPurdy: yup,

Issue #824, opened 2016-08-15
Font-smoothing is non-standard


revoltpuppy:
**What version of Sass Lint are you using?**
grunt-sass-lint: ^0.2.0

**Please include any relevant parts of your configuration**

<pre><code>no-vendor-prefixes:
    - 2
    -
      additional-identifiers:
        - khtml
        - o
      ignore-non-standard: true</code></pre>

**What did you expect to happen?**
Expected the linter to ignore `-webkit-font-smoothing` and `-moz-osx-font-smoothing`.

According to MDN, [font-smooth (not font-smoothing, by the way) is non-standard](https://developer.mozilla.org/en-US/docs/Web/CSS/font-smooth).

**What actually happened? Please include any error messages given to you by Sass Lint.**
Got error: “Vendor prefixes should not be used—no-vendor-prefixes” for both uses.

DanPurdy:
@revoltpuppy thanks for this. Will look to get it fixed soon

Nazin:
Any update on this issue?

mjyoung:
Also would love for this to be fixed.

alanshortis:
Another plus one for this. It's a bit of a blocker for me right now.

DanPurdy:
Sorry about the wait, had to take a bit of an unschedule hiatus for a while there.

We will be releasing a V2 of sass-lint in the next couple of weeks to mainly drop support for old node versions but also to replace the vendor prefix rule. I've moved rules similar to stylelint's approach to this over to sass-lint but with added customisation options! Hopefully should remove the issues you've all been having.

PR #1126, opened 2017-08-31
Fix issue with no-vendor-prefix rule & non-standard props

bgriffith:
This PR references the `known-css-properties` list to check against standard properties only when `ignore-non-standard` config setting is `false`.

```yml
no-vendor-prefixes:
    - 2
    -
      ignore-non-standard: true
```

Closes https://github.com/sasstools/sass-lint/issues/824
`<DCO 1.1 Signed-off-by: Ben Griffith gt11687@gmail.com>`

### DIFF
--- a/lib/rules/no-vendor-prefixes.js
+++ b/lib/rules/no-vendor-prefixes.js
@@ -1,12 +1,8 @@
 'use strict';
 
-var helpers = require('../helpers'),
-    yaml = require('js-yaml'),
-    fs = require('fs'),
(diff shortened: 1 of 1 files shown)


DanPurdy:
This isn't the right approach. I have a half finished implementation which uses autoprefixr to handle this for us. I mentioned to ya the other day that this was next in my list after handling the other property list rule.
The list you're referencing here includes the non standard properties and our list should be deprecated and removed.

PR #1127, opened 2017-09-02
Allow functions & custom props in rgba params

### LINKED ISSUE #853: Reporting color functions as invalid uses of color literals
Using `gulp-sass-lint` **v1.2.0**

---

I have this code

``` scss
$color-rgba: rgba(red($input-border-focus), green($input-border-focus), blue($input-border-focus), .6);
box-shadow: $form-shadows, 0 0 8px $color-rgba;
```

Which produces these errors, which seems incorrect since the color names above are [the built-in SCSS functions](http://sass-lang.com/documentation/Sass/Script/Functions.html#red-instance_method), **not** color literals.

```
  27:22  warning  Color functions such as 'red' should not be passed to functions, use variables    no-color-literals
  27:48  warning  Color functions such as 'green' should not be passed to functions, use variables  no-color-literals
  27:76  warning  Color functions such as 'blue' should not be passed to functions, use variables   no-color-literals
```

---

When re-written as a single line like this

``` scss
box-shadow: $form-shadows, 0 0 8px rgba(red($input-border-focus), green($input-border-focus), blue($input-border-focus), .6);;
```

I only get this one error instead

```
  27:39  warning  Color functions such as 'rgba' should only be used in variable declarations  no-color-literals
```

---

Unless I'm missing something this seems like a bug to me, and I don't see a way around this except to either disable this rule (which seems like a bad solution) or to just exclude this one file from linting right now.  

I was able to re-write this code into 4 variables, which is a little bit of an insane work around... but at least it accepts the code now

``` scss
//Due to a bug in `sass-lint` this must be broken up into multiple variables.
//Otherwise it sees the color function as color literals and reports them as violations :/
$color-r: red($input-border-focus);
$color-g: green($input-border-focus);
$color-b: blue($input-border-focus);
$color-rgba: rgba($color-r, $color-g, $color-b, .6);
box-shadow: $form-shadows, 0 0 8px $color-rgba;
```

DanPurdy:
Functions being passed as params of rgba were flaggins as literals. This change allows functions to be passed properly as params of rgba.

Fixes #853 

There's a lot of other things mentioned in that ticket but I believe all the others to be either misconfigured rules or invalid configs. Happy to keep the issue open if we don't agree though.

### DIFF
--- a/lib/rules/no-color-literals.js
+++ b/lib/rules/no-color-literals.js
@@ -80,7 +80,7 @@ module.exports = {
 
     ast.traverseByTypes(['value', 'variable'], function (node, i, parent) {
 
-      // If we don't literals as variable names then check each variable name
+      // If we don't allow literals as variable names then check each variable name
       if (node.is('variable') && !parser.options['allow-variable-identifiers']) {
         if (cssColors.indexOf(node.content[0].content) !== -1) {
           result = helpers.addUnique(result, {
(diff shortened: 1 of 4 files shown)


lib/rules/no-color-literals.js
```diff
@@ -127,15 +127,18 @@ module.exports = {
             // if rgba usage is allowed we need to make sure only variables are being passed to it.
             else if (
               parser.options['allow-rgba'] &&
-              funcType === 'rgba' &&
-              valElem.content[1].content[0].type !== 'variable' &&
+              funcType === 'rgba' && (
+                valElem.content[1].content[0].type !== 'variable' &&
```
bgriffith: Suggest we use `!valElem.content[1].content[0].is('variable')` etc

Issue #1051, opened 2017-03-22
Update regex for the hyphenatedbem option in Class Name Format


mikeriley131:
What version of Sass Lint are you using?
1.10.2

Please include any relevant parts of your configuration
```
class-name-format:
    - 1
    -
      'convention':
      - 'hyphenatedbem'
```
What did you do? Please include the actual source code causing the issue.
```
.footer {
  ...
}

.footer__contact-info--list {
  ...
}
```

What did you expect to happen?
Valid hyphenated BEM

What actually happened? Please include any error messages given to you by Sass Lint.
```
Warning	Class '.footer' should match regular expression /hyphenatedbem/	class-name-format
Warning	Class '.footer__contact-info--list' should match regular expression /hyphenatedbem/	class-name-format
```
I believe the regex should get updated to:
`^([\.\%]?[a-z]*[-]?[a-z0-9\-]*)(\.[a-z0-9\-]*)?(__[a-z0-9]*[-]?[a-z0-9\-]*)?(--[a-z0-9]*[-]?[a-z0-9\-]*)?(\:[a-z]*)*$`

as seen here: https://regex101.com/r/yP0eM7/1 (via https://github.com/brigade/scss-lint/issues/240)

DanPurdy:
Hey @mikeriley131 I think this is actually a config issue as your examples parse correctly as hyphenated BEM. It should be set as follows
```yaml
  class-name-format:
    - 1
    -
      convention: hyphenatedbem
```

The way you're specifying your option meant that the rule was actually trying to match the rule against a regex of `\hyphenatedbem\` so unless your classname was called `.hyphenatedbem` it would fail 😄 

I will update the rule with a slight variation of that regex though, as it does seem to be a bit more fit for purpose.

Thanks

DanPurdy:
On second thoughts that regex wont work.. It would still class `----` as valid hyphenated BEM..

PR #1130, opened 2017-09-02
Update hyphenated BEM regex pattern

DanPurdy:
related to #1051 

slight update to check for actual hyphenated BEM rather than just the characters that are allowed in the pattern. I've slightly modified what was suggested though as we don't need to look for `.` or `%` at the beginning of the name.


DanPurdy:
This is actually crap... nvm

Issue #952, opened 2016-11-14
class-name-format not working with variable interpolation


ncoden:
I am using `gulp-sass-lint v1.3.2` with `sass-lint v1.10.2`.

For the following code:
```scss
    // Static (unresponsive) row gutters
    @each $breakpoint, $value in $grid-column-gutter {
      &.#{$gutter}-#{$breakpoint} {
        > .#{$column} {
          @include grid-col-gutter($value);
        }
      }
    }
```
(from [foundation-sites / grid/_classes.scss](https://github.com/zurb/foundation-sites/blob/546a5b9f38487450df0dcf4957ba8c73bf11194e/scss/grid/_classes.scss#L59))

The following warning is throw on `&.#{$gutter}-#{$breakpoint}`:
```
   59:19  warning  Class '.-' should match regular expression /([a-z0-9]+-?)+/  class-name-format
```

The sass-lint config is :
```yml
...
  class-name-format:
    - 1
    - convention: '([a-z0-9]+-?)+'
...
```
(from [.sass-lint.yml](https://gist.github.com/ncoden/ec50bad6cd2f08df0780e9f356442794))

Sass-lint seems to ignore variable interpolation, and so consider the variable name as `.-`. I expect `class-name-format` to consider the class name as `\.(.*)\-(.*)` for the regex check.

DanPurdy:
I'll double check this, shouldn't be throwing any issues here.

coreyjv:
Is there an update on this bug?

ncoden:
Poke @DanPurdy

DanPurdy:
Sorry for this insane delay, life getting in the way of things... I will check this in #1131 as we should be ignoring interpolated classnames in the next release.

DanPurdy:
Yup, just run it in the tests this doesn't throw the error anymore. WIll mark it as fixed with that PR but give it a final test before we release it. Sorry again for the delay.

DanPurdy:
Released in 1.12.0

PR #1131, opened 2017-09-02
Feature/ignore interpolation class names

### LINKED ISSUE #988: fix string interpolation in class names during class-name-format validation
> What version of Sass Lint are you using?

1.9.1

> Please include any relevant parts of your configuration

```
rules:
  class-name-format: [2, { convention: strictbem } ]
```

> What did you do? Please include the actual source code causing the issue.

```scss
$primary-class: 'bla-bla-bla';

.#{$primary-class}--small {
  ...

  &__child {
    ...
  }
}
```

> What did you expect to happen?

No error.

> What actually happened? Please include any error messages given to you by Sass Lint.

[sass-lint] Class '.--small' should be written in BEM (Block Element Modifier) format
False positive error. Class name should be 'bla-bla-bla--small' and should pass BEM class name validation.

> If you're using a IDE plugin have you tried the CLI too?

IDE Plugin, VS Code

DanPurdy:
Fixes #988 

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -309,7 +309,10 @@ helpers.attemptTraversal = function (node, traversalPath) {
       currentNodeList = [],
       processChildNode = function processChildNode (child) {
         child.forEach(traversalPath[i], function (n) {
-          nextNodeList.push(n);
+          if (n.content && n.content.some && n.contains('interpolation')) {
+            return false;
+          }
+          return nextNodeList.push(n);
         });
       };
 
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -309,7 +309,7 @@ helpers.attemptTraversal = function (node, traversalPath) {
       currentNodeList = [],
       processChildNode = function processChildNode (child) {
         child.forEach(traversalPath[i], function (n) {
-          if (n.content && n.content.some && n.contains('interpolation')) {
+          if (n.content && typeof n.content !== 'string' && n.contains('interpolation')) {
             return false;
           }
           return nextNodeList.push(n);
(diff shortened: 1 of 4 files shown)


lib/helpers.js
```diff
@@ -309,7 +309,10 @@ helpers.attemptTraversal = function (node, traversalPath) {
       currentNodeList = [],
       processChildNode = function processChildNode (child) {
         child.forEach(traversalPath[i], function (n) {
-          nextNodeList.push(n);
+          if (n.content && n.content.some && n.contains('interpolation')) {
```
bgriffith: Are you using `some` here to determine if the array has length? Normally `Array.prototype.some` requires a callback function - if above suggest just using `n.content.length`

DanPurdy: Well content might be a string and a string still has a length property so that's not a good way to determine if it's an array. We're not invoking `some` just checking if it's defined on the prototype of whatever is in content and stops the rule crashing if it then tries to use the contain method which doesn't exist on a content of type string.

bgriffith: Why not just use isArray?

DanPurdy: Yeah i'll be honest i dont really know why i did it this way. I think I was actually doing something else with a `some` at the same time and decided to use it. odd.. will change it now anyways.

DanPurdy:
So best we can do for now is to use typeof i think and check we aren't trying to use against a string. When we update to v2 we can go ahead and use Array.isArray as it's in Node 4.5 lts. It's not in 0.10 or 0.12 which technically our v1 library supports. Even though im sure we've merged in some node >4.5 dependencies 😄 oops... Either way lets get this in to fix the issue, maybe gonzales should look to check it's error throwing as it's a gonzales crash that this prevents.

DanPurdy:
Fixes #952

PR #1026, opened 2017-02-27
Whitelist files published to npmjs - Closes #986.

JakeChampion:
Closes #986

**What do the changes you have made achieve?**
Limits the files published to npmjs to only those required to run `sass-lint`, not the ones needed to develop `sass-lint`.

**Are there any new warning messages?**
None have been added.

**Have you written tests?**
No tests have been added.

**Have you included relevant documentation**
Not sure what documentation would need to be updated.

**Which issues does this resolve?**
This reduces the size on `sass-lint` on npmjs from 15MB to 13MB. The majority of the size comes from:
```
 4.8M	node_modules/lodash
4.9M	node_modules/globule
```

`<DCO 1.1 Signed-off-by: jake champion me@jakechampion.name>`

PR #1132, opened 2017-09-02
Npm ignore

### LINKED ISSUE #986: Add to package.json whitelist files
Hi, 
sass-lint installation include unnecessary files (tests, .travis.yml and other). It increases the size and installation time.

npm has directive `files` in `package.json` [doc](https://docs.npmjs.com/files/package.json#files)

I created [example](https://github.com/ahiipsa/sass-lint/blob/files/package.json#L15-L21)

You may test it:
```
npm install git@github.com:ahiipsa/sass-lint.git#files
```

What do you think about this?

DanPurdy:
fixes #986 

Removes all files and directories not required to run sass-lint to minimise the install size.

To test you can checkout this branch and run `npm pack` it should create a tar file then do `tar -tf THE_FILE_NAME` to get a list of all the files that would be created when installing from NPM

See also https://docs.npmjs.com/misc/developers

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,14 @@
+.github
+coverage
+docs
+tests
+.editorconfig
+.eslintrc
+.gitignore
+.npmignore
+.npmrc
+.travis.yml
+.nvmrc
+appveyor.yml
+CONTRIBUTING.md
+CODE_OF_CONDUCT.md


DanPurdy:
Fixes #1026

Issue #735, opened 2016-05-31
Make formatters pluggable / Add formatter for VSO (aka TFS, VSTS)


mattbrooks2010:
## FEATURE REQUEST
Add a new formatter to support VSO / TFS integration.

## PROPOSED SOLUTION
Implement a new 'VSO' formatter, either built-in or as a plugin. Both options require changes to the way `sass-lint` is currently implemented (see status below).

The solution would be to emit [VSO logging commands](https://github.com/Microsoft/vsts-tasks/blob/master/docs/authoring/commands.md), which are similar to [TeamCity service messages](https://confluence.jetbrains.com/display/TCD9/Build+Script+Interaction+with+TeamCity#BuildScriptInteractionwithTeamCity-servMsgsServiceMessages).

## STATUS
Blocked because formatters are loaded directly from ESLint Core. The ESLint team are no longer  adding new formatters to ESLint Core.

Please see https://github.com/eslint/eslint/issues/6262.

DanPurdy:
Thanks @mattbrooks2010 I'm not sure why they were directly using the eslint ones, I think it was a stop gap to get support for formatters before and it's just never been updated since I've been on board 😕 You're right though this definitely needs rectifying soon.

WIll look into it.

mattbrooks2010:
The ESLint team are no longer adding new formatters directly to ESLint Core.

It may therefore be worth making a simple change to [sassLint.format](https://github.com/sasstools/sass-lint/blob/fbaf73825dfe57bc6827d1d4bd79bba513fd5934/index.js#L243) to:

1. Retain support for existing ESLint formatters and existing user configurations, at least in the short term.
2. Allow additional formatters to be developed as standalone packages and used in an intuitive manner.

My suggestion is to hardcode the names of the [14 existing ESLint formatters](https://github.com/eslint/eslint/tree/master/lib/formatters) to support (1) above and establish a new convention for referencing formatter packages for (2).

A couple of alternatives for (2) based on the existing code would be:

```js
// Load formatter packages using formatter name directly
var formatted = require(config.options.formatter);
return formatted(results);

// Load formatter packages by applying a convention to the formatter name
var formatted = require(`sass-lint-formatter-${config.options.formatter}`);
return formatted(results);

// Load formatter packages by applying an alternative convention
var formatted = require(`sass-lint-${config.options.formatter}`);
return formatted(results);
```

Thoughts?

webdevnerdstuff:
[...]

PR #1134, opened 2017-09-03
Port eslint formatters & custom formatter loader

### LINKED ISSUE #521: incorrect link in html report to rule description
I am running sass-lint from the command line with

```
sass-lint --verbose --format html  --syntax scss --config ~/.sass-lint.yml my.scss > ~/report.html

```

the config file is the one in
https://github.com/sasstools/sass-lint/blob/master/lib/config/sass-lint.yml

the report.html file generated has links to rule descriptions that are not found e.g.

no-ids : http://eslint.org/docs/rules/no-ids

no-vendor-prefixes : http://eslint.org/docs/rules/no-vendor-prefixes

is there some configuration that I am missing ?

### LINKED ISSUE #1052: absolute path to formatter fails
sass-lint v1.10.2

passing --format /some/formatter/index.js fails because sass-lint will pass that through to eslint, which means eslint will take as eslint/lib/formatters//some/formatter/index.js.

Judging from the docs, sass-lint should call getFormatter in eslint to resolve that to a formatter.

### LINKED ISSUE #1070: ESLint as dependency?
I had a brief look at the source code, and don't find that `eslint` is being used in the source itself, seems like it's only being used for linting the source code and not as a utility library. May I know why it's listed under `dependencies` in `package.json`?

If it's just for linting the source code, it would be more appropriate to be under `devDependencies`. Putting an outdated `eslint` as dependency causes projects to not be able to upgrade to the latest `eslint` which is > `3.0.0` at the moment.

Would be happy to submit a PR to fix this if necessary. Please let me know (:

DanPurdy:
We've been depending upon the eslint formatters for a long time now, this causes a few issues.
* We become dependent on a eslint verison which in v1 blocked a lot of our users from using us as they had to use a more recent eslint
* All eslint specific mentions in formatters couldn't be customised to point at our docs etc
* We allowed all formats that eslint offered even if they made no sense to us

I've ported across most of the eslint formatters and tests (why reinvent the wheel?) with correct credits and license information. A few of the formatters I don't think made sense or were very specifically js focused so they've not moved.

This PR also gives a user the ability to specify a path to a custom formatter too so in theory any formatters we don't port a user can just point at their eslint install anyways.

Some of the tests look for a fatal flag too which we currently don't (but should!) support for things where a fatal error is thrown. This will need a bit more core work in sass-lint to achieve but definitely something we should look to achieve.

Tasks
- [x]  Port formatters
- [x] Port tests
- [x] Port format loader
- [x] Port license
- [x] Add documentation
- [x] Update cwd logic to specify a cwd in config

Fixes #1052 
helps out with #735 
fixes #521 
fixes #1070 


Zelig880:
Hi Guys,

Are this merge conflict going to be merged any time soon?  Do we know the turnaround for this kind of tasks

DanPurdy:
@Zelig880 I'm hoping to get this done in the next week or so, time is scarce at the moment. This will form a v2 update as we're dropping support for a few things and change a few bits and pieces around. I would expect us to start working on v3 straight away once this is out as we want to make quite a few breaking changes but obviously holding them and releasing in one go is annoying for everyone.

PR #1137, opened 2017-09-03
Add new vendor prefix rules

### LINKED ISSUE #824: Font-smoothing is non-standard

### LINKED ISSUE #843: ignore-non-standard does not ignore -webkit-tap-highlight-color
1.9.0

``` YAML
rules:
  no-vendor-prefixes:
    - 2
    -
      ignore-non-standard: true
```

``` SCSS
html {
  -webkit-tap-highlight-color: $link-color-hover;
}

button::-moz-focus-inner,
input::-moz-focus-inner {
  border: 0;
  padding: 0;
}

input[type="number"]::-webkit-inner-spin-button,
input[type="number"]::-webkit-outer-spin-button {
  height: auto;
}
```

I ran the no-vendor-prefixes rule on the example for `ignore-non-standard` which then caused an error. If the `ignore-non-standard` config is set to false, 5 errors are thrown which is expected.

What did you expect to happen?
The -webkit-tap-highlight-color property to be ignored when `ignore-non-standard: true`.

What actually happened? Please include any error messages given to you by Sass Lint.
An error was thrown.
The -webkit-tap-highlight-color property caused an error to be thrown.
`2:3    Error   Vendor prefixes should not be used  no-vendor-prefixes`

If you're using a IDE plugin have you tried the CLI too?
The above occurred using the CLI.

### LINKED ISSUE #868: Idea: add a ignore-properties option to the no-vendor-prefixes rule
I am using version `1.9.1` and am currently affected by issue #843 regarding the `ignore-non-standard` option not working, but that gave me an idea for a new option/setting to add to the `no-vendor-prefixes` rule.

Right now when`ignore-non-standard` is `true` it just ignores all properties that aren't listed in the [standard properties list](https://github.com/sasstools/sass-lint/blob/master/data/properties.yml), but IMO this can be too leniant in some cases.  Instead I'd love to have the ability to just list out the non-standard properties that I want to allow in my application, instead of just allowing them all.

An example configuration for this might look like this

``` yml
  no-vendor-prefixes:
    - 1
    -
      'ignore-properties':
        - '-ms-overflow-style'
        - '-ms-touch-action'
        - '-webkit-font-smoothing'
        - '-webkit-tap-highlight-color'
        - '-webkit-overflow-scrolling'
        -'::-webkit-scrollbar'
        -'::-webkit-scrollbar-button'
        -'::-webkit-scrollbar-track'
        -'::-webkit-scrollbar-thumb'
```

If this option were added my application could ONLY use these vendor prefixes, and not any others.  Since so many developers at my organization might touch this code, this really helps limit the kinds of styles they might add.  Since we use `autoprefixer` it's not needed for a developer to add any vendor prefixes except for the instances listed above.

### LINKED ISSUE #886: no-vendor-prefixes -ms-clear problem
**What version of Sass Lint are you using?**
1.9.1

**Please include any relevant parts of your configuration**

```
  no-vendor-prefixes:
   - 2
   - ignore-non-standard: true
```

**What did you do? Please include the actual source code causing the issue.**

```
    &::-ms-clear {
      display: none;
      width : 0;
      height: 0;
    }
```

**What did you expect to happen?**
No errors

**What actually happened? Please include any error messages given to you by Sass Lint.**
error  Vendor prefixes should not be used                                   no-vendor-prefixes

**If you're using a IDE plugin have you tried the CLI too?**
That's from CLI

### LINKED ISSUE #953: no-vendor-prefixes confuse standard properties and pseudo-elements
I am using `gulp-sass-lint v1.3.2` with `sass-lint v1.10.2`.

For the following code:
```scss
progress {
    &::-ms-fill { ... }
}
```
(from [foundation-sites / forms/_progress.scss](https://github.com/zurb/foundation-sites/blob/546a5b9f38487450df0dcf4957ba8c73bf11194e/scss/forms/_progress.scss#L86))

The following warning is thrown:
```
  86:8  warning  Vendor prefixes should not be used  no-vendor-prefixes
```

The sass-lint config is :
```yml
...
  no-vendor-prefixes:
     - 1
     - ignore-non-standard: true
...
```
(from [.sass-lint.yml](https://gist.github.com/ncoden/ec50bad6cd2f08df0780e9f356442794))

Sass-lint seems (I'm not sure about it) to consider the progress pseudo-element `::-ms-fill` as _standard_ because of the `fill` property. But `::-ms-fill` is specific to IE10 (see https://developer.mozilla.org/fr/docs/Web/CSS/::-ms-fill). I expect `ignore-non-standard` to not check for a standard _pseudo-element_ in a list of _property_.

### LINKED ISSUE #1143: some non-standard properties are in standard properties list.
```
"sass-lint": "^1.10.2",
```
```
  no-vendor-prefixes:
    - 1
    - ignore-non-standard: true
```
```
.text {
  font-size: 12px;
  overflow: hidden;
  text-overflow: ellipsis;
  display: -webkit-box;
  -webkit-line-clamp: 2;
  -webkit-box-orient: vertical;
  -webkit-tap-highlight-color: transparent;
}
```
and then got three warnning: 
```
warning  Vendor prefixes should not be used  no-vendor-prefixes
```

What did you expect to happen?

**no warning for these properties only supported with the prefix or non-standard.**

- line-clamp https://caniuse.com/#search=line-clamp
- box-orient which is not a standard property now. https://developer.mozilla.org/en-US/docs/Web/CSS/box-orient
- tap-highlight-color: not a standard property either. https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-tap-highlight-color

And also if write them without prefix, autoprefixer neither prefix them.

DanPurdy:
Adds 5 new rules
* no-selector-vendor-prefix
* no-property-vendor-prefix
* no-value-vendor-prefix
* no-at-rule-vendor-prefix
* no-media-feature-vendor-prefix

Taken a leaf out of the stylelint book here as their rule was a lot more mature than ours and with the number of problems our property list was causing it made sense to break these rules up. I've added the ability to add additional properties to lint and exclude from linting too as I felt this covered all possible scenarios our users could now have for these rules.

The rules use Autoprefixer to say whether a property should be prefixed or not. If Autoprefixer would handle that then it will flag an error on the 'thing' it would normally prefix itself. This does obviously make the assumption that people are using Autoprefixer in their pipeline but it also does mean it may be a little soft on some rules which is where the exclude properties option will come in handy.

I also updated our data lists for pseudo elements and classes with the latest from MDN (we should look to move these out too at some point). 

The property sort order rule was the last rule using our css properties list so i've updated that too to use the new list and removed the properties list from the repo.

The old `no-vendor-prefixes` rule has also been removed. 

This will be a V2 release

fixes #953
fixes #824 
fixes #843
fixes #886 
closes #868 
fixes #1143


DanPurdy:
I need to remove the old vendor prefix rule from the config file

IrisYao:
Great! So when will have a new release?  Thanks

PR #1139, opened 2017-09-14
chore: fixes and updates dev dependencies on nyc/eslint

JaKXz:
- moves from Istanbul to nyc
- moves eslint to devdeps

ATM, `eslint` is pulled down as a dependency of `sass-lint` when installing via npm or yarn. I don't think this is necessary, and is causing dependency resolution issues in my projects.

Additionally, `istanbul` has moved on to the `nyc` command line tool: https://github.com/istanbuljs/nyc where the underlying libraries have been improved.

`<DCO 1.1 Signed-off-by: Jason Kurian jasonk92@gmail.com>`


DanPurdy:
Sorry mate this isn't acceptable. We are already in the process of moving eslint out of our dependency chain but we can;t right now because the current sass-lint relies on the formatters directly. You also can't go changing the node versions we test against right now as that will inevitably mean a major version bump. Again we're moving to the LTS versions of Node in our v2 update. As for Istanbul thanks for the reminder, you'll see this update very soon.

Issue #1019, opened 2017-02-14
Possible to disallow initial value?


Donaldini:
Is it possible to disallow the use of certain values?

My specific use case is the use of the _initial_ value inside e.g. the _color_ property.
I want to disable this because this value is [not supported by IE](http://caniuse.com/#feat=css-initial-value), but seems like an excellent fix for some issues (until you test in IE).

`color: initial // Does not work on IE`

**What version of Sass Lint are you using?** 1.10.2

DanPurdy:
There's a PR open in #1142 which will add a rule to do this.

agnoster:
Though notably #1142 only disallows values for a specific property - a blanket blacklist for all instances of the use of the `initial` value across all property names isn't possible. But it could be extended to allow e.g.

```yaml
no-disallowed-property-values:
  properties:
    '*': initial
```

PR #1142, opened 2017-09-21
New rule: no-disallowed-property-values

agnoster:
NOTE: I haven't done the DCO/Emoji stuff yet because I'm not sure if this is a desirable addition. If it is, I'll be happy to make any additional changes necessary to merge.

This should allow blacklisting certain property **values**, in contrast to no-disallowed-properties which blacklists the property **names**. For example, this allows banning `text-transform: capitalize` while allowing other uses of `text-transform`.

Just the ones from the lint rule.

**Have you written tests?**

I have added tests based off the tests for `no-disallowed-properties`.

I have added a new doc in `docs/rules/no-disallowed-property-values.md`

No open issues that I found.

`<DCO 1.1 Signed-off-by: Isaac Wolkerstorfer isaac@asana.com>`


lib/rules/no-disallowed-property-values.js
```diff
@@ -0,0 +1,45 @@
...
+      var propertyNameNode = propertyNode.first();
+      if (!propertyNameNode.is('ident')) {
+        return;
+      }
+      var propertyName = propertyNameNode.content;
+      var disallowedPropertyValues = parser.options.properties[propertyName];
+      if (!disallowedPropertyValues) {
+        return;
+      }
+      if (typeof disallowedPropertyValues === 'string') {
```
DanPurdy: You allow property values to be specified as strings outside of arrays here, you don't have this in your tests, it would be good to add this case in. Maybe duplicate the ` display: ['inline'] ` as just `display: 'inline'` that would help.

docs/rules/no-disallowed-property-values.md
```diff
@@ -0,0 +1,30 @@
...
+
+The following would not be allowed:
+
+```scss
+
+// the capitalize value for text-transform is not allowed
+.foo {
+  text-transform: capitalize;
+}
+
```
DanPurdy: Can we just add a quick example of multiple property values also, just to be complete in how we allow the property lists to be defined.

DanPurdy:
closes #1019

agnoster:
Great comments! Updated the docs and tests for those cases - I think we actually don't need the "array with one value" case if we're testing "array with two values" and "single string value"?

lib/rules/no-disallowed-property-values.js
agnoster: Done! Though I ended up just replacing the `display: ['inline']` case with `display: 'inline'` which feels like it should be sufficient?

docs/rules/no-disallowed-property-values.md
agnoster: Done!

DanPurdy:
Hi @agnoster it would be good to include both. so single value as a string, single and multiple values in an array. It makes sense to test all the allowed cases like this just so we cover all cases. If you wouldn't mind just adding that case too then we could get this merged 👍

agnoster:
Ah, fair enough, good reasoning! Done :-D

timothympace:
What's the status of this PR? I'm hoping to use this rule in a project.

Also, is there any plan to support blacklisting values that are strings? I noticed that if a property value is set to a string, `ast.traverseByType('value', function (valueNode, idx, parent) {` will not pick it up. String values have a node type of 'string'. And will there be any support for blacklisting values that are resolved through a variable?

agnoster:
Looks like it got approved but never merged?

DanPurdy:
Sorry, our develop branch is now work for v2 and I just haven't had time to go back through everything and merge it all. I will hopefully in the next few days and release this. It will be part of our v2 update though which just removed the eslint dependencies etc.

PR #1144, opened 2017-09-22
Add updated CSS properties to Concentric sorting

### LINKED ISSUE #964: Concentric: transition properties separated
1.10.2

```coffee
'property-sort-order': [
  1
  order: 'concentric'
]
```

With code like the following...
```scss
.some-class {
  transition: right 0.4s;
  transition-timing-function: ease-in-out;
  min-height: 500px;
}
```

...sass-lint complains that `min-height` and `transition-timing-function` should be swapped. This seems bizarre and counterintuitive.

In https://github.com/sasstools/sass-lint/blob/develop/lib/config/property-sort-orders/concentric.yml, many related properties are explicitly grouped (`flex` with `flex-*`, `animation` with `animation-*`, etc.), but `transition` properties are not.

It seems all possible `transition` properties should be grouped together for consistency.

Since the concentric property order seems to derive from https://github.com/brigade/scss-lint/blob/master/data/property-sort-orders/concentric.txt, I wasn't sure whether to post the issue here or with `scss-lint`.

### LINKED ISSUE #1105: Add missing transform-* properties in concentric sort order
There are some css properties missing in the concentric.yml

- transform-box
- transform-origin
- transform-style

sass-lint/lib/config/property-sort-orders/concentric.yml

I suggest adding them here : 
```
  - 'transform'
  - 'transform-box'
  - 'transform-origin'
  - 'transform-style'
  - 'transition'
```

occupant:
Adds missing CSS properties to Concentric sorting, including:

- box-sizing
- all grid properties
- additional transform properties
- additional transition properties
- counter properties
- page-break properties

Updated order to reflect most recent version of concentric ordering at: https://github.com/brandon-rhodes/Concentric-CSS/blob/master/style3.css

Fixes #964 
Fixes #1105 

`<DCO 1.1 Signed-off-by: James O'Toole james.otoole@ubc.ca>`

### DIFF
--- a/lib/config/property-sort-orders/concentric.yml
+++ b/lib/config/property-sort-orders/concentric.yml
@@ -2,8 +2,13 @@
 # http://rhodesmill.org/brandon/2011/concentric-css/
 
 # Based on: https://github.com/brigade/scss-lint/blob/master/data/property-sort-orders/concentric.txt
+# Updated list at: https://github.com/brandon-rhodes/Concentric-CSS/blob/master/style3.css
 
 order:
+  # box-sizing
+  - 'box-sizing'
+  
+  # position
   - 'display'
   - 'position'
   - 'top'
@@ -11,6 +16,15 @@ order:
   - 'bottom'
   - 'left'
 
+  - 'float'
+  - 'clear'
+
+  # align-content
+  - 'align-content'
+  - 'align-items'
+  - 'align-self'
+
+  # flex
   - 'flex'
   - 'flex-basis'
   - 'flex-direction'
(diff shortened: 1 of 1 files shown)


DanPurdy:
Awesome, thanks for this. Im debating with myself whether this would constitute a breaking change or not. Basicaly we're planning to do one last v1 update, drop some node version support in v2 and then work on v3 all in the space of a few weeks. I'll merge this in the next couple of days though, thanks again! 👍

DanPurdy:
Ok, It's going to be a breaking change so will have to come out with the v2 due to the existing properties being re arranged. Will merge this into our V2 branch and hopefully you will see it live soon. Thanks!

PR #1165, opened 2017-12-22
Replace `front-matter` with `gray-matter`

### LINKED ISSUE #1164: Replace `front-matter` with `gray-matter`
Replace `front-matter` with `gray-matter`.

**Why?**

`gray-matter` does not use regular expressions to match front-matter. As a result, it's [faster](https://github.com/jonschlinkert/gray-matter/blob/master/benchmark/stats.md) and more flexible than `front-matter`

Regarding flexibility, since gray-matter doesn't use regex for matching, it has other features that might be useful (even if they aren't, gray-matter is still faster and more performant):

- [stringifying](https://github.com/jonschlinkert/gray-matter#stringify) back to front matter from an object
- jekyll-style [excerpts](https://github.com/jonschlinkert/gray-matter#optionsexcerpt)
- [custom delimiters](https://github.com/jonschlinkert/gray-matter#optionsdelimiters) 
- [custom engines](https://github.com/jonschlinkert/gray-matter#optionsengines)

doowb:
Replace `front-matter` with `gray-matter`.

**Why?**

`gray-matter` does not use regular expressions to match front-matter. As a result, it's [faster](https://github.com/jonschlinkert/gray-matter/blob/master/benchmark/stats.md) and more flexible than `front-matter`

Regarding flexibility, since gray-matter doesn't use regex for matching, it has other features that might be useful (even if they aren't, gray-matter is still faster and more performant):

- [stringifying](https://github.com/jonschlinkert/gray-matter#stringify) back to front matter from an object
- jekyll-style [excerpts](https://github.com/jonschlinkert/gray-matter#optionsexcerpt)
- [custom delimiters](https://github.com/jonschlinkert/gray-matter#optionsdelimiters) 
- [custom engines](https://github.com/jonschlinkert/gray-matter#optionsengines)

Closes #1164 

`<DCO 1.1 Signed-off-by: Brian Woodward <brian.woodward@gmail.com>>`


DanPurdy:
Nice one, Thank you for this.

doowb:
No problem! Thank you!

PR #1166, opened 2017-12-26
Update smacss sort order

VladShcherbin:

The `smacss` order is updated to the latest version. This version was [approved](https://github.com/brigade/scss-lint/pull/884) earlier this year and is used in other packages.

Probably not.

**Have you written tests?**

No.

Not needed.

None.

`<DCO 1.1 Signed-off-by: Vlad Shcherbin vlad.shcherbin@gmail.com>`

### DIFF
--- a/lib/config/property-sort-orders/smacss.yml
+++ b/lib/config/property-sort-orders/smacss.yml
@@ -5,14 +5,24 @@
 
 order:
 
+  # Heading
+
+  - 'content'
+  - 'quotes'
+
   # Box
 
   - 'display'
+  - 'visibility'
+
   - 'position'
+  - 'z-index'
   - 'top'
   - 'right'
   - 'bottom'
   - 'left'
+
+  - 'box-sizing'
 
   - 'flex'
   - 'flex-basis'
@@ -30,7 +40,6 @@ order:
   - 'width'
   - 'min-width'
   - 'max-width'
-
   - 'height'
   - 'min-height'
   - 'max-height'
@@ -50,6 +59,13 @@ order:
   - 'float'
   - 'clear'
 
+  - 'overflow'
+  - 'overflow-x'
+  - 'overflow-y'
+
+  - 'clip'
+  - 'zoom'
+
   - 'columns'
   - 'column-gap'
   - 'column-fill'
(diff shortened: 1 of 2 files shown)


DanPurdy:
Do you think there's a reason not to call this a new sort order? I mean i can see from the PR to scss-lint that it's been updated there. I'm just thinking that maybe this should be smacss2017 or something? Happy to hear opinions @VladShcherbin as i don't personally ever use sort orders.

VladShcherbin:
@DanPurdy not sure if a new preset is needed.

I actually found out about new rules when tried `stylelint` with `stylelint-config-property-sort-order-smacss` preset, which uses [css-property-sort-order-smacss](https://github.com/cahamilton/css-property-sort-order-smacss) package inside. This package has updated syntax.

Personally, I'm happy with sass-lint and I thought that it would be great to have the same syntax everywhere (in scss-lint, sass-lint, stylelint, etc) so a user can easily migrate from one to another.

Since new rules [were approved](https://github.com/brigade/scss-lint/pull/884#issuecomment-277683235) by `smacss` author, I think it'll be fine to update and stick to one style in all projects.

DanPurdy:
Perfect, thanks for the explanatation @VladShcherbin might be a good idea to move over to the package at some point then so this can be updated easier.

Going to be releasing a sass-lint v2 at some point soon for some breaking changes we've made, hoping to get it out soon.

VladShcherbin:
@DanPurdy any chance to release 2.0 alpha or some version of 1.* with this?

JWess:
Any idea when this might release?

vladshcherbin:
It would be great to release updated order as 1.*.0 so it could be used until 2.0 is released, seems like it takes more time then expected :)

@JWess meanwhile, you can use the updated version with this command (for yarn):

```console
yarn add sass-lint#6a71e29920a386fd97b986eb54ed0123890c1913 -D
```

JWess:
@vladshcherbin Thanks for the update. I also noticed that sass-lint is again behind scss-lint, as scss-lint has now added the grid properties:

https://github.com/brigade/scss-lint/blob/master/data/property-sort-orders/smacss.txt

PR #1167, opened 2017-12-28
Update dependencies

realityking:
Bumps `fs-extra` and `known-css-properties` to their newest releases.


DanPurdy:
Hmm, i wonder why greenkeeper hasn't been doing these. Thanks @realityking

PR #1172, opened 2018-01-09
Update smacss property order

### LINKED ISSUE #1171: Using property-sort-order with order: smacss applies an out-of-date property list
When the property-sort-order rule is used with pre-canned order "smacss", the property order that is applied comes from smacss.yml, which is based upon https://github.com/brigade/scss-lint/blob/master/data/property-sort-orders/smacss.txt. That file has undergone a number of revisions in recent years and smacss.yml is now out-of-date. The changes include some adjustments of order and grouping, plus inclusion of CSS animation, list-style and grid properties.

To maintain the usefulness of this option we should update the property order list in smacss.yml to the latest version.

dcwarwick:
Update the list in smacss.yml to match https://github.com/brigade/scss-lint/blob/master/data/property-sort-orders/smacss.txt. This list also included 'transform-box', 'background-attachment', 'background-clip' and 'font-smoothing', so these are retained in the natural places.

This ensures that the property list and sort order applied when using the property-sort-order rule with order: smacss will use the latest sort order and include all relevant properties.

This PR includes an update to the tests, as the test file now produces 14 warnings when the smacss sort order is applied. It does not include documentation as it is a straight update of an existing behaviour.

Closes #1171 

`<DCO 1.1 Signed-off-by: Dave Clark dave.clark@uk.ibm.com>`

### DIFF
--- a/lib/config/property-sort-orders/smacss.yml
+++ b/lib/config/property-sort-orders/smacss.yml
@@ -5,14 +5,49 @@
 
 order:
 
+  # Heading
+  
+  - 'content'
+  - 'quotes'
+
   # Box
 
   - 'display'
+  - 'visibility'
+  
   - 'position'
+  - 'z-index'
   - 'top'
   - 'right'
   - 'bottom'
   - 'left'
+  
+  - 'box-sizing'
+  
+  - 'grid'
+  - 'grid-after'
+  - 'grid-area'
+  - 'grid-auto-columns'
+  - 'grid-auto-flow'
+  - 'grid-auto-rows'
+  - 'grid-before'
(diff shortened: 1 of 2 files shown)


VladShcherbin:
@dcwarwick https://github.com/sasstools/sass-lint/pull/1166, although I pulled it in master 🚚

DanPurdy:
Yeah, that was my bad for not checking the branch before i merged it, oops...

oh well this will be updated once we push the release out anyways which hopefully will be soon!

JWess:
This PR included the new grid properties, while #1166 did not. We need a PR to include the grid props.

PR #1230, opened 2018-11-03
Bring smacss sort order configuration up to date with scss-lint's

ccjmne:
Quoting the lastest comment (@JWess's) in #1166:

> ... `sass-lint` is behind `scss-lint`, as `scss-lint` has now added the grid properties:
> 
> https://github.com/brigade/scss-lint/blob/master/data/property-sort-orders/smacss.txt

I updated the `smacss` sort order file according to the reference.

Update `smacss` sort order to match that of other projects.

Shouldn't be – will check out later on.

**Have you written tests?**

Yes. I updated the tests accordingly, just has it was done the last time we brought `smacss.yaml` up to date. The tests update are within their own separate commit (a7ce9ae).

Deemed irrelevant.

#1171 (sort of), once published. I didn't create an issue just for this.

### DIFF
--- a/lib/config/property-sort-orders/smacss.yml
+++ b/lib/config/property-sort-orders/smacss.yml
@@ -5,14 +5,49 @@
 
 order:
 
+  # Heading
+
+  - 'content'
+  - 'quotes'
+
   # Box
 
   - 'display'
+  - 'visibility'
+
   - 'position'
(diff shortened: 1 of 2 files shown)


JWess:
This still hasn't released. Is there an ETA?

PR #1233, opened 2018-11-10
Update to 1.13.0

chancestrickland:


DanPurdy:
If this gets to be released it needs to be v2 im afraid - there are breaking changes

